PR #60, opened 2018-08-07
Do not show python version incompatible results in dashboard

liyanhui1228:


grid_builder.py
```diff
@@ -49,6 +49,17 @@ def __init__(self, package_to_results, pairwise_to_results):
         self._package_to_results = package_to_results
         self._pairwise_to_results = pairwise_to_results
 
+    def _is_py_version_incompatible(self, result):
+        if result.status == compatibility_store.Status.INSTALL_ERROR:
+            for version in [2, 3]:
+                for pkg in result.packages:
```
ylil93: I think the contained if statement may have too many things going on inside of it, maybe you can split things up for example like this:

```
                for pkg in result.packages:
                    major_version = result.python_major_version
                    name = pkg.install_name
                    unsupported = configs.PKG_PY_VERSION_NOT_SUPPORTED[version]
                    if major_version == version and name in unsupported:
                        return True
```

liyanhui1228: Done, thx!

PR #67, opened 2018-08-15
Add system test for compatibility checker server

liyanhui1228:
For #55

### DIFF
--- a/system_test/test_compatibility_checker_server.py
+++ b/system_test/test_compatibility_checker_server.py
@@ -17,7 +17,6 @@ import os
 import requests
 import signal
 import subprocess
-import time
 
 from retrying import retry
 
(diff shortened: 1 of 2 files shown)


system_test/test_compatibility_checker_server.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class TestCompatibilityCheckerServer(unittest.TestCase):
+
+    def setUp(self):
+        # Run application
+        self.process = run_application()
+
+        # Wait app to start
+        time.sleep(3)
```
wangxf123456: Will this make tests flaky? Are there some apis to check whether the app is ready?

liyanhui1228: Good catch! Added retry for waiting the app to start.

PR #69, opened 2018-08-16
System test for badge_server

### LINKED ISSUE #68: System test for badge server

liyanhui1228:
Closes #68 
MemoryStore (Redis) can only be accessed within the Kubernetes cluster, so we set the environment variable to indicate run locally in the system test, and patch away the redis client methods.


badge_server/badge_server.py
```diff
@@ -67,6 +67,28 @@
...
+
+
+def fake_redis_set(*args, **kwargs):
+    key = args[2][0]
+    value = str(args[2][1]).encode('utf-8')
+    REDIS_CACHE[key] = value
+
+
+# Patch away the redis connections if run locally
+if os.environ.get('RUN_LOCALLY') is not None:
```
wangxf123456: Is this only used for test? If it is, why don't we put this in the test cases because it is unrelated to the functionalities of the badge server?

system_test/test_badge_server.py
```diff
@@ -0,0 +1,113 @@
...
+import unittest
+
+HOST_PORT = '0.0.0.0:8080'
+BASE_URL = 'http://0.0.0.0:8080/'
+PACKAGE_FOR_TEST = 'opencensus'
+
+RETRY_WAIT_PERIOD = 8000 # Wait 8 seconds between each retry
+RETRY_MAX_ATTEMPT = 10 # Retry 10 times
+
+EXPECTED_SVG = b'<svg xmlns="http://www.w3.org/2000/svg" xmlns:' \
```
wangxf123456: This looks scary... Should we put this in a separate file?

badge_server/badge_server.py
liyanhui1228: Because the `wrapt` magic only works in the same process, and in the test we start a separate process to run the badge server. So if we put that logic in tests, the redis client won't be patched away.

system_test/test_badge_server.py
liyanhui1228: Done.

PR #82, opened 2018-08-27
Adding dependency badging

ylil93:
also refactored dependency highlighter for usage in badge server

### DIFF
--- a/badge_server/badge_server.py
+++ b/badge_server/badge_server.py
@@ -170,27 +170,47 @@ def _get_pair_status_for_packages(pkg_sets):
     return version_and_res
 
 
-def _get_badge_url(version_and_res, package_name):
-    # By default use the status of py3, if it is not SUCCESS, and it can be
-    # installed with python2, then use the result of python2.
+#def _get_badge_url(version_and_res, package_name):
+#    # By default use the status of py3, if it is not SUCCESS, and it can be
+#    # installed with python2, then use the result of python2.
+#    package_name = package_name.replace('-', '.')
+#    status = version_and_res['py3']['status']
+#    if status != 'SUCCESS' and \
+#            package_name not in configs.PKG_PY_VERSION_NOT_SUPPORTED.get(2):
+#        status = version_and_res['py2']['status']
+#
+#    color = STATUS_COLOR_MAPPING[status]
+#    url = URL_PREFIX + '{}-{}-{}.svg'.format(
+#        package_name, status, color)
+#
+#    return url
+#
+#
+#def _get_dep_badge_url(res, package_name):
+#    package_name = package_name.replace('-', '.')
+#    status = res['status']
+#
+#    color = DEP_STATUS_COLOR_MAPPING[status]
+#    url = URL_PREFIX + '{}-{}-{}.svg'.format(
+#        package_name, status, color)
+#
+#    return url
+
+
+def _get_badge_url(res, package_name):
     package_name = package_name.replace('-', '.')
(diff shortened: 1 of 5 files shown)


compatibility_lib/compatibility_lib/dependency_highlighter.py
```diff
@@ -34,17 +34,17 @@ class UnstableReleaseError(Exception):
 
 
 class PriorityLevel(enum.Enum):
-    NOT_SET = "UP_TO_DATE"
-    LOW = "LOW_PRIORITY"
-    HIGH = "HIGH_PRIORITY"
+    UP_TO_DATE = 0
+    LOW = 1
+    HIGH = 2
```
ylil93: Should I preserve "LOW_PRIORITY" and "HIGH_PRIORITY" as the enum names?

badge_server/badge_server.py
```diff
@@ -120,6 +123,20 @@ def fake_redis_set(*args, **kwargs):
...
+    priority_level.LOW.name:          'black',
+    priority_level.HIGH.name:         'red'
+}
+
+DEP_CONVERSION_ERROR_RES = {
+    'status': 'CONVERSION_ERROR',
+    'details': None,
+}
+
+
```
liyanhui1228: Only one blank line is needed here.

badge_server/badge_server.py
```diff
@@ -170,6 +187,17 @@ def _get_badge_url(version_and_res, package_name):
     return url
 
 
+def _get_dep_badge_url(res, package_name):
```
liyanhui1228: This could be combined with `_get_badge_url` above.

badge_server/badge_server.py
```diff
@@ -279,6 +307,75 @@ def self_compatibility_badge_target():
...
+
+    package_name = flask.request.args.get('package')
+
+    res = {
+        'status': 'CALCULATING',
+        'details': None,
+    }
+
+    def run_check():
+        # First see if this package is already stored in BigQuery.
```
liyanhui1228: Remove this comment? Looks like accidentally copied.

compatibility_lib/compatibility_lib/dependency_highlighter.py
liyanhui1228: Yes that's good.

badge_server/badge_server.py
```diff
@@ -279,6 +307,75 @@ def self_compatibility_badge_target():
...
+        outdated = highlighter.check_package(package_name)
+
+        max_level = priority_level.UP_TO_DATE
+        details = []
+        for dep in outdated:
+            level = dep.priority.level
+            if level.value > max_level.value:
```
liyanhui1228: Why having this? The details could just be `None` if the version is up to date.

ylil93: this snippet does 2 things:
1. track the highest outdated dependency (to be displayed on the badge)
2. append detail string for every outdated dependency to the list

if the list is empty, that means no outdated dependencies and the package is up to date and details is `EMPTY_DETAILS`

if the list is not empty, the list of strings are joined together to produce one string and provides info on all dependencies that are outdated

liyanhui1228: Sounds good, thanks for the explanation! And please also reply `done` or any comment you prefer to the rest of the review comments in future PRs, which makes it easier to track which were resolved. Thanks!

ylil93: ok! thanks for the review!

PR #89, opened 2018-08-29
Make dashboard display all the check warnings

### LINKED ISSUE #86: Dashboard only shows one of the compatibility check message
In the table cell, if there are multiple reason causing the compatibility check to fail, currently it only display one of them. As we now have the dialog to display the details, we should show all of the failing details.

liyanhui1228:
Fix #86


dashboard/grid_builder.py
```diff
@@ -115,37 +130,63 @@ def get_result(self,
...
-            return {
-                'status': compatibility_store.Status.SUCCESS.name,
-                'self': False,
-            }
+                )
+                if cell_color_status is 'Self-SUCCESS':
+                    cell_color_status = 'Pairwise-SUCCESS'
+
+        result = {
+            'color_status': cell_color_status,
```
brianquinlan: Are you sure that you want to split the UI logic between the Python code and the template?

dashboard/grid_builder.py
```diff
@@ -101,12 +109,19 @@ def get_result(self,
                             May be None.>
             }
         """
+        self_result = []
+        pair_result = []
+        cell_color_status = 'Self-SUCCESS'
```
brianquinlan: does it make sense to always use the lowercase name to make the names easier to grep for, etc.?

dashboard/grid_builder.py
```diff
@@ -70,13 +72,19 @@ def _is_py_version_incompatible(self, result):
...
+        # Get self result
         for package_2 in self._package_to_results.keys():
-            result = self.get_result(p, package_2)
+            results = self.get_result(p, package_2)
             # Don't report the package as having issues if it is purely the
             # result of a self-incompatibility of another package.
-            if result['status'] != compatibility_store.Status.SUCCESS.name and (
-                    not result['self'] or p == package_2):
-                return True
+            pure_self_conflict = False
```
brianquinlan: What does "pure_self_conflict" mean?

dashboard/grid_builder.py
liyanhui1228: Done.

dashboard/grid_builder.py
liyanhui1228: It means the pairwise check failure is because one of the packages is not conflict with itself. And it enables us to differentiate the two types of pairwise failures in the dashboard as described below:
e.g google-cloud-dataflow and tensorflow is failing with status color purple(self conflict), while google-cloud-dataflow and google-cloud-language is failing with status color orange(pairwise conflict).

dashboard/grid_builder.py
liyanhui1228: Could you explain a little more here?

dashboard/grid_builder.py
```diff
@@ -70,13 +72,19 @@ def _is_py_version_incompatible(self, result):
 
     def has_issues(self, p: package.Package) -> bool:
         """Returns true if the given package has any compatibility issues."""
+        # Get self result
         for package_2 in self._package_to_results.keys():
-            result = self.get_result(p, package_2)
+            results = self.get_result(p, package_2)
```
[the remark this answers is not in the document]
liyanhui1228: Done.

dashboard/grid_builder.py
```diff
@@ -70,13 +72,19 @@ def _is_py_version_incompatible(self, result):
...
             # Don't report the package as having issues if it is purely the
             # result of a self-incompatibility of another package.
-            if result['status'] != compatibility_store.Status.SUCCESS.name and (
-                    not result['self'] or p == package_2):
-                return True
+            pure_self_conflict = False
+            self_res = self.get_result(package_2, package_2)
```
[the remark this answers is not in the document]
liyanhui1228: Done.

dashboard/grid_builder.py
liyanhui1228: Renamed `cell_color_status` to `status_type`. Avoid mentioning the template specific stuff in `grid_builder.py`.

dashboard/grid_builder.py
liyanhui1228: Done.

PR #90, opened 2018-09-06
improve test case readability

ylil93:


compatibility_lib/compatibility_lib/test_dependency_highlighter.py
```diff
@@ -164,46 +165,29 @@ def setup_test__get_update_priority(self):
...
-            priority(level.HIGH_PRIORITY, major_version),
-            eval(ptemp % ((2,5,0)+(3,0,4)+(1,)))
-        ))
-
-        cases.append((
-            priority(level.HIGH_PRIORITY, major_version),
-            eval(ptemp % ((2,5,0)+(5,0,4)+(1,)))
-        ))
+
+        def dictify(args):
```
liyanhui1228: Can we make this args explicit by passing the major, minor, patch into this method?

compatibility_lib/compatibility_lib/test_dependency_highlighter.py
```diff
@@ -146,8 +146,9 @@ def setUp(self):
             self._store)
 
     def setup_test__get_update_priority(self):
-        priority = dependency_highlighter.Priority
-        level = dependency_highlighter.PriorityLevel
+        LOW = dependency_highlighter.PriorityLevel.LOW_PRIORITY
```
liyanhui1228: We don't usually use upper case variable name inside functions.

ylil93: done

compatibility_lib/compatibility_lib/test_dependency_highlighter.py
ylil93: I let `run()` take in `install_tup` and `latest_tup` so that when you call that function, you can group the major, minor, and patch of the different versions and I think it's more straightforward than creating a dict on the fly. `run((2,5,0), (2,6,0), 5)`

And since they were passed in as tuples, I think it's more elegant to let `dictify()` take in a tuple as the argument otherwise the unpackaging can get messy.

Let me know if you disagree with any of that.

PR #92, opened 2018-09-10
Display the packages that having deprecated dependencies

### LINKED ISSUE #88: Display the deprecated dependency for each package in the dashboard

liyanhui1228:
Fixes #88


dashboard/grid-template.html
```diff
@@ -117,7 +117,9 @@
     {% set row = loop.index0 %}
     <tr>
         <td><font face="verdana">{{ row_package.friendly_name }}</font>{% if results.has_issues(row_package) %}
-                <img class="warning-img" src="https://www.gstatic.com/images/icons/material/system/1x/warning_amber_18dp.png">
+                {% set issues = results.deprecated_deps[row_package.friendly_name][0] %}
+                <img class="warning-img" class="dialog" onclick="showIssues('{{ row_package.friendly_name|safe }}', '{{ issues|replace('\'', '\\\'')|safe }}')" src="https://www.gstatic.com/images/icons/material/system/1x/warning_amber_18dp.png">
```
brianquinlan: Can't jinja2 do this escaping for you? I'm not sure.

dashboard/grid_builder.py
```diff
@@ -88,8 +96,32 @@ def has_issues(self, p: package.Package) -> bool:
...
+    def get_deprecated_deps(self):
+        """
+        Returns True if there are deprecated dependencies for a
+        given package.
+        """
+        finder = deprecated_dep_finder.DeprecatedDepFinder()
+        deprecated_deps = list(finder.get_deprecated_deps())
+
+        results = {}
+        for item in deprecated_deps:
```
brianquinlan: Could you use tuple unpacking so that item[0][X] has a name?

dashboard/grid_builder.py
```diff
@@ -88,8 +96,32 @@ def has_issues(self, p: package.Package) -> bool:
...
+
+            # Whether the package has deprecated dependencies or not
+            if self.deprecated_deps[p.install_name][1]:
+                return True
+
         return False
 
+    def get_deprecated_deps(self):
+        """
+        Returns True if there are deprecated dependencies for a
```
brianquinlan: Maybe add the type annotation? This doesn't return a bool.

dashboard/grid_builder.py
liyanhui1228: Done.

dashboard/grid-template.html
liyanhui1228: Removed, jinja2 couldn't do the escaping for `'`, so I have to manually escape it here.

dashboard/grid_builder.py
liyanhui1228: Done.

Issue #85, opened 2018-08-28
Dependency Highlighter marks pip and setuptools as high priority for every package


liyanhui1228:
Looks like `pip` and `setuptools` should be treated specially.

Things to do:
- Investigate what's the stable version of `pip` and `setuptools` is.
- If the stable (or widely used version by every package) is not the latest, we should have it in the config file and do not mark it as high priority.

ylil93:
I noticed that those packages were not up to date as requirements in many google cloud packages, but can I ask if there's any other reason they need to be treated specially?

What I found so far:
pip:
pinned version = 10.0.1
https://blog.python.org/2018/04/pip-10-has-been-released.html
https://github.com/pypa/pip/issues/5221

setuptools:
pinned version = 39.0.1
The only thing that stuck out to me is that version 40.0.0 drops support for Python 3.3. 
https://setuptools.readthedocs.io/en/latest/history.html

liyanhui1228:
We would need to treat them specially because they are not direct dependencies for most of the Google owned packages, and the package owners can't do anything to resolve the high priority warning on `pip` and `setuptools` as they are not installed via pip. (We aren't able to `pip install` anything before we install `pip`)

And the best way to go probably is to filter out these two dependencies when analyzing the outdated versions.

PR #101, opened 2018-09-14
special handling for pip and setuptools (#85)

ylil93:


compatibility_lib/compatibility_lib/configs.py
```diff
@@ -14,6 +14,11 @@
 
 """Common configs for compatibility_lib."""
 
+IGNORED_DEPENDENCIES = [
```
liyanhui1228: Add a comment about why we ignore these dependencies would be nice.

ylil93: Done.

PR #142, opened 2018-10-04
Support one badge usage

### LINKED ISSUE #126: Support single badge usage
Adding single badge to github repo readme file, clicking on it will display the package detail page.

liyanhui1228:
Fixes #126 

This also changes the usage of calling badge server endpoint:
`.../image/...` -> `..._image/...`

### DIFF
--- a/badge_server/badge_server.py
+++ b/badge_server/badge_server.py
@@ -287,8 +287,10 @@ def index():
 @app.route('/one_badge_image')
 def one_badge_image():
     package_name = flask.request.args.get('package')
+    # Remove the last '/' from the url root
     url_prefix = flask.request.url_root[:-1]
-    # Call the url for each badge to run the checks.
+    # Call the url for each badge to run the checks. This will populate the
+    # individual caches, which are used to calculate the final image state.
     # Self compatibility badge
     requests.get(url_prefix + flask.url_for(
         'self_compatibility_badge_image', package=package_name))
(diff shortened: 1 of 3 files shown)


badge_server/badge_server.py
```diff
@@ -222,7 +284,44 @@ def index():
     return 'Visitor number: {}'.format(value)
 
 
-@app.route('/self_compatibility_badge/image')
+@app.route('/one_badge_image')
+def one_badge_image():
+    package_name = flask.request.args.get('package')
+    url_prefix = flask.request.url_root[:-1]
+    # Call the url for each badge to run the checks.
```
brianquinlan: ... the checks. This will populate the individual caches, which are used to calculate the final image state.

badge_server/badge_server.py
```diff
@@ -222,7 +284,44 @@ def index():
...
+    color = STATUS_COLOR_MAPPING[status]
+    package_name = package_name.replace('-', '.')
+    url = URL_PREFIX + '{}-{}-{}.svg'.format(package_name, status, color)
+
+    return requests.get(url).text
+
+
+@app.route('/one_badge_target')
+def one_badge_target():
+    package_name = flask.request.args.get('package')
```
brianquinlan: Don't you have to populate the cache here?

ylil93:
Looks good to me, but I'm not knowledgeable about flask, so I'll defer to Brian

badge_server/badge_server.py
liyanhui1228: This route is for generating the package detail page, which uses the cached result calculated by calling `one_badge_image`. We don't need to populate the cache here. So this must run after `one_badge_image` is called (and github badges works like this), or the package page will show an empty status.

badge_server/badge_server.py
liyanhui1228: Done.

PR #154, opened 2018-10-10
google-cloud-dataflow renamed to apache-beam[gcp]

liyanhui1228:
`google-cloud-dataflow` stops releasing new versions, and changed to `apache-beam[gcp]`.


liyanhui1228:
We will need to do a new release of the compatibility_lib and run a new kokoro build to update the data in bigquery after this change.

PR #156, opened 2018-10-10
Remove the work around for renaming

liyanhui1228:
Will merge after the kokoro build finished running.


liyanhui1228:
The tests should pass after the build.

PR #157, opened 2018-10-11
Use the latest version for the dependencies

liyanhui1228:
In py3, `pip install apache-beam` will install 2.2.0 version; In py2, it will install 2.7.0.

Changed to use the latest version for storing into the dependency info table.

### DIFF
--- a/compatibility_lib/compatibility_lib/compatibility_store.py
+++ b/compatibility_lib/compatibility_lib/compatibility_store.py
@@ -428,8 +428,7 @@ class CompatibilityStore:
             return True
 
         install_name = cs.packages[0].install_name
-        install_name_sanitized = install_name.split('[')[0] \
-            if '[' in install_name else install_name
+        install_name_sanitized = install_name.split('[')[0]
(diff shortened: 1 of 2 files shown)


compatibility_lib/compatibility_lib/compatibility_store.py
```diff
@@ -416,6 +422,26 @@ def save_compatibility_statuses(
...
                 row)
 
+    def _should_update_dep_info(self, cs, dep_info_stored):
+        """Return True if the stored version is behind latest version."""
+        if dep_info_stored is None:
+            return True
+
+        install_name = cs.packages[0].install_name
+        install_name_sanitized = install_name.split('[')[0] \
+            if '[' in install_name else install_name
```
ylil93: if there is no '[' in the `install_name`, `install_name.split('[')` will return a list of size 1 containing `install_name`, so I think this line is redundant

liyanhui1228: Removed.

Issue #163, opened 2018-10-15
Compatibility server should allow the user to inject a name into badge left-hand-side


brianquinlan:
Like:

?name=head&...  =>
   "up-to-date dependencies (head): UP TO DATE"

?name=pybadges&...  =>
   "cross compatibility(pybadges): SUCCESS"

PR #164, opened 2018-10-16
Make the badge name configurable

### LINKED ISSUE #162: Badges should have names other than the package name
How about:

"Self Compatibility" => "internal compatibility"
"Google Compatibility" => "cross compatibility"
"Dependency Version Status" => "up-to-date dependencies"

liyanhui1228:
Fixes #162 and #163


badge_server/badge_server.py
```diff
@@ -334,8 +337,7 @@ def one_badge_image():
 
     status, _, _, _ = _get_all_results_from_cache(package_name)
     color = STATUS_COLOR_MAPPING[status]
-    package_name = _sanitize_package_name(package_name)
```
ylil93: can you explain why _sanitize_package_name() is only called if badge_name is None instead of every time?

liyanhui1228: Good catch! I was thinking if badge_name is None, use the package_name as the badge_name. But it could be possible that the user specified badge_name also needs to be sanitized. Fixed.

PR #174, opened 2018-11-02
Update readme with the app engine endpoint

liyanhui1228:


badge_server/README.rst
```diff
@@ -1,8 +1,6 @@
 Badge Server
 ============
 
-.. image:: http://35.226.8.89/one_badge_image?package=tensorflow
-
```
ylil93: Is this supposed to be replaced with the new endpoint?

liyanhui1228: I felt like this was added accidentally from previous PR. So removing it...

PR #181, opened 2018-11-07
Docker in docker for compatibility server

liyanhui1228:
This PR implements the docker in docker logic to run the compatibility check. It does the things below:
- Used docker python sdk (https://pypi.org/project/docker/) to create/run/commit container image inside the application docker image.
- Generated and modified the deployment yaml file to bind mount /var/run/docker.sock, which equals to the docker command `docker run -v /var/run/docker.sock:/var/run/docker.sock -p 8888:8888 gcr.io/xxx/compatibility-server:v1`.

This workflow won't require restarting servers after each request, cause the installed packages are inside the internal docker, which won't interfere with each other. But the latency grows up in this new workflow. I used opencensus trace to instrument the latency and result as below:
screenshot from 2018-11-06 16-41-28

And for previous restart workflow:
screenshot from 2018-11-06 16-42-23

The latency increased as we need to build the container in each step, but this should be tolerable as we don't expect user to directly use this server.


compatibility_server/pip_checker.py
```diff
@@ -197,20 +199,54 @@ def __init__(self,
...
+        container = self._docker_client.containers.run(
+            container_name,
+            command=command,
+            stderr=stderr,
+            detach=detach)
+        result = container.wait()
+        returncode = result.get('StatusCode')
+        return container, returncode
+
+    def _run_command(self, command, stdout_path, stderr_path):
```
brianquinlan: Your approach of saving and restoring the docker container is too clever ;-)

I think that it would be simpler (and possibly much faster) to do something like (this is just conceptual):

```python

def run()
  self._container = self._docker_client.containers.run(
            container_name,
            command='sleep 300', # exit if not done in 5 minutes, that's all of our CPU that you can use for bitcoin ;-)
            detach=True)
  # More stuff

def _run_command(command, stdout_path, stderr_path, raise_on_failure=True):
  exit_code, _ = self._container.exec_run(command)
  ...
```

liyanhui1228: Good suggestion! I replaced the container commit with exec_run and it's much faster!

liyanhui1228: screenshot from 2018-11-07 13-11-09

liyanhui1228:
Circle CI runs all the tests on a docker container itself, and we will need to mount the volume of it to run the system tests. Still looking into how to config circle ci to mount volumes...

compatibility_server/pip_checker.py
```diff
@@ -230,74 +254,54 @@ def _call_pypi_json_api(pkg_name, pkg_version):
     def _build_command(self, subcommands):
         return self._pip_command + subcommands
 
-    def _freeze(self, requirements_file_path):
+    def _freeze(self):
```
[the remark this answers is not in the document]
liyanhui1228: Removed.

compatibility_server/pip_checker.py
```diff
@@ -230,74 +254,54 @@ def _call_pypi_json_api(pkg_name, pkg_version):
...
         command = self._build_command(['freeze'])
-        std_err_path = os.path.join(self._output_directory, 'freeze-error.txt')
-        self._run_command(command, requirements_file_path, std_err_path)
+        return self._run_command(command, stdout=True, stderr=False)
 
     def _uninstall(self, requirements_file_path):
```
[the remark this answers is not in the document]
liyanhui1228: Removed.

compatibility_server/pip_checker.py
```diff
@@ -197,20 +199,42 @@ def __init__(self,
...
+
+    def _cleanup_container(self):
+        self._container.stop()
+        self._container.remove()
 
-        logging.debug('Running %s [returncode=%s]', command,
-                      completed_command.returncode)
-        if completed_command.returncode and raise_on_failure:
-            raise PipError(command, completed_command.returncode, stderr_path)
+    def _run_command(self, command, stdout, stderr, raise_on_failure=True):
```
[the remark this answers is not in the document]
liyanhui1228: Added a docstring describing the types. The names are matching the docker python sdk, so that we can directly pass them into the api function call.

compatibility_server/pip_checker.py
```diff
@@ -197,20 +199,42 @@ def __init__(self,
...
-                                                         'w') as stderr_file:
-            completed_command = subprocess.run(
-                command, stdout=stdout_file, stderr=stderr_file)
+    def _get_base_image(self):
+        python_version = self._pip_command[0]
+
+        if python_version == 'python2':
+            base_image = "python:2.7"
+        else:
+            base_image = "python:3.6"
```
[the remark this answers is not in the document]
liyanhui1228: It has a complete Python interpreter and pip.
https://github.com/docker-library/python/blob/bf1acb4f1caad419ff290d700044240b4e8cb0df/3.6/stretch/Dockerfile

compatibility_server/pip_checker.py
```diff
@@ -197,20 +199,42 @@ def __init__(self,
         self._tmp_path = tmp_path
         self._output_directory = None
         self._clean = clean
+        self._docker_client = docker.from_env()
```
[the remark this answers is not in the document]
liyanhui1228: The `_container` is needed for most of the functions in this class, so I'd like to create it in the constructor rather that making a `self._container` in the `run` function, which will also work but looks a little strange.

compatibility_server/pip_checker.py
```diff
@@ -197,20 +199,42 @@ def __init__(self,
...
+        self._container = self._build_container
```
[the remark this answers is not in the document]
liyanhui1228: The `_build_container` has a `@property` decorator so that it's called as a parameter.

compatibility_server/pip_checker.py
```diff
@@ -354,12 +358,15 @@ def _list(self):
 
     def run(self):
         """Run the version compatibility check."""
-        self._output_directory = tempfile.mkdtemp(dir=self._tmp_path)
         if self._clean:
```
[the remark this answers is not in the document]
liyanhui1228: Removed.

compatibility_server/pip_checker.py
```diff
@@ -197,20 +199,54 @@ def __init__(self,
...
-        logging.debug('Running %s [returncode=%s]', command,
-                      completed_command.returncode)
-        if completed_command.returncode and raise_on_failure:
-            raise PipError(command, completed_command.returncode, stderr_path)
-
-        return completed_command.returncode
+        self._docker_client = docker.from_env()
+
+    def _get_base_image(self):
+        python_version = self._pip_command[0]
```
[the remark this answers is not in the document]
liyanhui1228: Removed.

compatibility_server/Dockerfile
```diff
@@ -68,29 +68,11 @@ RUN make distclean >/dev/null
 WORKDIR /tmp
 RUN rm -rf /tmp/Pyt/tmp/Python-3.6.5
 
-# Install an independent a pristine version of Python 2.7 for dependency
```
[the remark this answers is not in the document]
liyanhui1228: Updated the dockerfile to use `python:3.6` as the base image, and removed the install python interpreter steps.

compatibility_server/deployment/deploy.yaml
```diff
@@ -0,0 +1,55 @@
...
+        terminationMessagePath: /dev/termination-log
+        terminationMessagePolicy: File
+        volumeMounts:
+        - mountPath: /var/run/docker.sock
+          name: docker-sock
+      dnsPolicy: ClusterFirst
+      restartPolicy: Always
+      schedulerName: default-scheduler
+      securityContext: {}
+      terminationGracePeriodSeconds: 30
```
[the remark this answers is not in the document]
liyanhui1228: Just looked for the documentation for this field, "this field is for allowing the Pod to shut down gracefully by waiting for connections to terminate after a Pod to scheduled for deletion". And it seems like it's not related to the timeout for dependency check.

compatibility_server/Dockerfile
```diff
@@ -20,86 +20,19 @@
 #  - python 3.6.5: used for python3 pip installation (build from source)
 #  - python 2.7.15: used for python2 pip installation (build from source)
 
-FROM ubuntu:18.04
-
-WORKDIR "/tmp"
+FROM python:3.6
 
 # Create /etc/timezone so the package `tzdata` won't prompt for time zone information.
 ENV TZ=America/Los_Angeles
```
brianquinlan: The timezone setup shouldn't be necessary either.

compatibility_server/deployment/deploy.yaml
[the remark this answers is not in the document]
brianquinlan: That's what I meant - if your requests take XX seconds then this value should be >XX seconds to allow requests to finish before the pod is shut down.

compatibility_server/pip_checker.py
```diff
@@ -27,29 +27,37 @@
...
 
+import docker
+
 PYPI_URL = 'https://pypi.org/pypi/'
+CONTAINER_WITH_PKG = "checker"
 
 
-class PipError(Exception):
-    """A pip command failed in an unexpected way."""
+class PipCheckerError(Exception):
```
brianquinlan: Maybe you should have PipCheckerError which just takes an error string (for use in the docker failure case) and a subclass, PipError, which takes the command.

compatibility_server/pip_checker.py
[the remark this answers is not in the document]
brianquinlan: I don't think that you want attribute lookups to have side effects like this.

compatibility_server/pip_checker.py
```diff
@@ -177,43 +185,83 @@ class _OneshotPipCheck():
...
-        logging.debug('Running %s [returncode=%s]', command,
-                      completed_command.returncode)
-        if completed_command.returncode and raise_on_failure:
-            raise PipError(command, completed_command.returncode, stderr_path)
+    def _run_command(self, command, stdout, stderr, raise_on_failure=True):
+        """Run docker commands using docker python sdk.
 
-        return completed_command.returncode
+        Args:
+            command (str): The command to run in docker container.
```
brianquinlan: Why not just use type annotations? That is done elsewhere in the file.

compatibility_server/pip_checker.py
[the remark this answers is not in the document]
brianquinlan: It's up to you. I try to avoid using external dependencies in constructors because it can make testing difficult i.e. because you can't even create the object without invoking the external dependency.

But I can see how your way is cleaner.

compatibility_server/deployment/deploy.yaml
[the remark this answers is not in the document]
liyanhui1228: Got it! I added timeout for dependency check and changed this field to be greater than the timeout.

compatibility_server/pip_checker.py

compatibility_server/pip_checker.py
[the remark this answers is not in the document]
liyanhui1228: Moved to `run`.

compatibility_server/pip_checker.py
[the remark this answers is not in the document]
liyanhui1228: Sounds good! Moved the docker stuff to `run`.

liyanhui1228:
PTAL

compatibility_server/deployment/deploy.yaml
```diff
@@ -0,0 +1,55 @@
...
+metadata:
+  annotations:
+    deployment.kubernetes.io/revision: "3"
+  creationTimestamp: 2018-06-15T20:08:48Z
+  generation: 8
+  labels:
+    run: compatibility-server
+  name: compatibility-server
+  namespace: default
+  resourceVersion: "19908269"
```
wangxf123456: Where do these values from? For example, "creationTimestamp", "resourceVersion", "uid".

compatibility_server/deployment/deploy.yaml
```diff
@@ -0,0 +1,55 @@
...
+        terminationMessagePath: /dev/termination-log
+        terminationMessagePolicy: File
+        volumeMounts:
+        - mountPath: /var/run/docker.sock
+          name: docker-sock
+      dnsPolicy: ClusterFirst
+      restartPolicy: Always
+      schedulerName: default-scheduler
+      securityContext: {}
+      terminationGracePeriodSeconds: 310
```
wangxf123456: Is this related to the 300 seconds timeout of _OneshotPipCheck._build_container()?

compatibility_server/compatibility_checker_server.py
```diff
@@ -71,30 +74,16 @@ def _parse_python_version_to_interpreter_mapping(s):
...
         """
         self._host = host
         self._port = port
-        self._clean = clean
-        self._python_version_to_interpreter = python_version_to_interpreter
-        self._install_once = install_once
-
-    def _shutdown(self):
-        threading.Thread(target=self._httpd.shutdown).start()
+        self._python_version_to_interpreter = PYTHON_VERSION_TO_INTERPRETER
```
brianquinlan: Do you need self. _python_version_to_interpreter ?

compatibility_server/compatibility_checker_server.py
```diff
@@ -124,24 +113,17 @@ def _check(self, start_response, python_version, packages):
...
         python_command = self._python_version_to_interpreter[python_version]
 
-        if self._install_once:
-            self._shutdown()
-
         try:
             pip_result = pip_checker.check(
-                [python_command, '-m', 'pip'], packages, clean=self._clean)
-        except pip_checker.PipError as pip_error:
+                [python_command, '-m', 'pip'], packages)
```
brianquinlan: I would still use an enum and not pass the whole command.

compatibility_server/pip_checker.py
```diff
@@ -27,29 +27,54 @@
...
+            'logs are: {error_msg}'.format(error_msg=error_msg))
 
-class PipError(Exception):
+        # Convert error_msg to bytes as this needed to be returned in the
+        # response which should be bytes. And the message returned from docker
+        # is already bytes so that won't need the conversion.
+        if isinstance(error_msg, str):
+            self.error_msg = error_msg.encode('utf-8')
+        else:
+            self.error_msg = error_msg
```
brianquinlan: Having an exception with an error message in bytes in odd.  You should convert from str to bytes at the HTTP boundary.

compatibility_server/pip_checker.py
```diff
@@ -177,43 +202,100 @@ class _OneshotPipCheck():
...
+        if python_version == 'python2':
+            base_image = "python:2.7"
+        else:
+            base_image = "python:3.6"
 
-        return completed_command.returncode
+        return base_image
+
+    def _build_container(self, docker_client):
+        """Build the container which contains a Python interpreter. The timeout
```
brianquinlan: First line of the command has to be a single sentence that fits on one line.

compatibility_server/pip_checker.py
```diff
@@ -177,43 +202,100 @@ class _OneshotPipCheck():
...
+        base_image = self._get_base_image()
+
+        try:
+            container = docker_client.containers.run(
+                base_image,
+                command="sleep {}".format(TIME_OUT),
+                detach=True)
+        except docker.errors.APIError as e:
+            raise PipCheckerError(
+                error_msg="Error occurs when starting docker container."
```
brianquinlan:
```suggestion
                error_msg="An error occurred while starting a docker container."
```

compatibility_server/pip_checker.py
```diff
@@ -177,43 +202,100 @@ class _OneshotPipCheck():
...
+
+    def _run_command(self,
+                     container: docker.models.containers.Container,
+                     command: List[str],
+                     stdout: bool,
+                     stderr: bool,
+                     raise_on_failure: Optional[bool] = True):
+        """Run docker commands using docker python sdk.
+
+        Args:
```
brianquinlan: You aren't documenting `container`.

1. Command is type annotated as List[str] while you document it as (str) in the argument documentation.
2. You don't need to duplicate the type documentation.

compatibility_server/pip_checker.py
```diff
@@ -177,43 +202,100 @@ class _OneshotPipCheck():
...
+            raise PipCheckerError(
+                error_msg="Error occurs when cleaning up docker container."
+                          "Container does not exist.")
+
+    def _run_command(self,
+                     container: docker.models.containers.Container,
+                     command: List[str],
+                     stdout: bool,
+                     stderr: bool,
+                     raise_on_failure: Optional[bool] = True):
```
brianquinlan: You can add the return value as a type annotation.

compatibility_server/pip_checker.py
```diff
@@ -177,43 +202,100 @@ class _OneshotPipCheck():
...
+        """Run docker commands using docker python sdk.
+
+        Args:
+            command (str): The command to run in docker container.
+            stdout (boolean): Whether to include stdout in output.
+            stderr (boolean): Whether to include stderr in output.
+            raise_on_failure (boolean): Whether to raise on failure.
+
+        Returns:
+            A tuple containing returncode and output.
```
brianquinlan: A
```suggestion
            A tuple containing the return code of the given command and the output of that command.
            Contains one or both of stdout/stderr depending on the values of `stdout` and `stderr`.
```

compatibility_server/pip_checker.py
```diff
@@ -177,43 +202,100 @@ class _OneshotPipCheck():
...
+        """
+        try:
+            returncode, output = container.exec_run(
+                command, stdout=stdout, stderr=stderr)
+
+            if isinstance(output, bytes):
```
brianquinlan: Is this sometimes bytes and sometimes a string?

compatibility_server/test_pip_checker.py
```diff
@@ -21,19 +21,120 @@
...
 import mock
 import os.path
+import subprocess
 import unittest
 
 import pip_checker
 
 
+def timestamp_to_seconds(timestamp):
+    """Convert a timestamp string into a microseconds value
```
brianquinlan: Inconsistent doc string style.

compatibility_server/test_pip_checker.py
```diff
@@ -21,19 +21,120 @@
...
+    :param timestamp
+    :return time in microseconds
+    """
+    from datetime import datetime
```
brianquinlan: Imports should be at the top level and should only include modules or packages.

liyanhui1228: Done.

compatibility_server/test_pip_checker.py
liyanhui1228: Done.

compatibility_server/pip_checker.py
liyanhui1228: Just confirmed everything returned from the internal docker is bytes.

compatibility_server/pip_checker.py

compatibility_server/pip_checker.py

compatibility_server/pip_checker.py
liyanhui1228: Fixed.

compatibility_server/pip_checker.py

compatibility_server/pip_checker.py
liyanhui1228: Removed the conversion.

compatibility_server/compatibility_checker_server.py
liyanhui1228: I changed to use a dict mapping the python version to the pip command. Seems like using an enum will add more conversions in here? Like `'2' -> PythonVersionEnum.PYTHON2 -> ['python2', '-m', 'pip']`

compatibility_server/compatibility_checker_server.py
liyanhui1228: Removed.

compatibility_server/pip_checker.py
```diff
@@ -177,43 +196,101 @@ class _OneshotPipCheck():
...
+                error_msg="Error occurs when cleaning up docker container."
+                          "Container does not exist.")
+
+    def _run_command(self,
+                     container: docker.models.containers.Container,
+                     command: List[str],
+                     stdout: bool,
+                     stderr: bool,
+                     raise_on_failure: Optional[bool] = True)\
+            -> (int, str):
```
brianquinlan: I think that you can move this to the previous line.

compatibility_server/pip_checker.py
```diff
@@ -27,29 +27,48 @@
...
-import subprocess
-import tempfile
+import socket
 import urllib.request
 
 from typing import Any, List, Mapping, Optional
 
+import docker
+
+socket.setdefaulttimeout(3)
```
brianquinlan: This is a bit of a code smell since it affects socket timeouts outside of the module. Since this server is the only place that this is used it will work but it might be better to use explicit timeouts when fetching URLs, if possible.

compatibility_server/pip_checker.py
```diff
@@ -27,29 +27,48 @@
...
-class PipError(Exception):
+        self.error_msg = error_msg
+
+
+class PipError(PipCheckerError):
     """A pip command failed in an unexpected way."""
 
-    def __init__(self, command: List[str], returncode: int, stderr_path: str):
+    def __init__(self,
+                 error_msg: List[str],
```
brianquinlan:
```suggestion
                 error_msg: str,
```

compatibility_server/pip_checker.py
```diff
@@ -27,29 +27,48 @@
...
         super(PipError, self).__init__(
-            'pip command ({command}) failed with error [{returncode}], '
-            'logs at: {stderr_path}'.format(
+            'Pip command ({command}) failed with error [{returncode}], '
+            'logs at: {error_msg}'.format(
                 command=command, returncode=returncode,
-                stderr_path=stderr_path))
+                error_msg=error_msg))
+
+        self.error_msg = error_msg
```
brianquinlan: I think that the base class will do this.

compatibility_server/test_pip_checker.py
```diff
@@ -18,22 +18,126 @@
...
 
 
+def timestamp_to_seconds(timestamp):
+    """Convert a timestamp string into a microseconds value
+    
+    Args:
+        timestamp: A timestamp string
+
+    Returns:
+        A string of the timestamp in microseconds.
```
brianquinlan: Does this really return a string?

compatibility_server/test_pip_checker.py
```diff
@@ -18,22 +18,126 @@
...
 import unittest
 
 import pip_checker
 
 
+def timestamp_to_seconds(timestamp):
+    """Convert a timestamp string into a microseconds value
+    
+    Args:
+        timestamp: A timestamp string
```
brianquinlan: Isn't this a  ISO 8601 string but without timezone support?

compatibility_server/pip_checker.py
liyanhui1228: Moved.

compatibility_server/pip_checker.py
liyanhui1228: Removed. Timeout is also handled in the internal docker.

compatibility_server/pip_checker.py

compatibility_server/pip_checker.py
liyanhui1228: Removed.

compatibility_server/test_pip_checker.py
liyanhui1228: Fixed.

compatibility_server/test_pip_checker.py
liyanhui1228: Modified.

compatibility_server/deployment/deploy.yaml
brianquinlan: You could add a comment to explain that.

compatibility_server/pip_checker.py
```diff
@@ -27,29 +27,44 @@
...
 
-    def __init__(self, command: List[str], returncode: int, stderr_path: str):
+    def __init__(self,
+                 error_msg: str,
+                 command: List[str],
+                 returncode: int):
         super(PipError, self).__init__(
-            'pip command ({command}) failed with error [{returncode}], '
-            'logs at: {stderr_path}'.format(
+            'Pip command ({command}) failed with error [{returncode}], '
```
brianquinlan:
```suggestion
            'Pip command ({command}) failed with error [{returncode}]: '
```

compatibility_server/pip_checker.py
```diff
@@ -27,29 +27,44 @@
...
+            'logs at: {error_msg}'.format(
```
brianquinlan:
```suggestion
            '{error_msg}'.format(
```

Because error_msg isn't a file path anymore.

compatibility_server/test_pip_checker.py
```diff
@@ -18,22 +18,127 @@
...
 
 import pip_checker
 
 
+def timestamp_to_seconds(timestamp: str) -> int:
+    """Convert a timestamp string into a seconds value
+    
+    Args:
+        timestamp: An ISO 8601 format string returned by calling isoformat()
+                   on a `datetime.datetime` type timestamp.
```
brianquinlan: Incorrect indentation style.

compatibility_server/pip_checker.py

compatibility_server/pip_checker.py

compatibility_server/test_pip_checker.py
liyanhui1228: Done.

PR #183, opened 2018-11-13
use a strict whitelist for sanitizing packages

ylil93:
The whitelist is hardcoded and contains pypi packages as well as github repo urls.


compatibility_server/test_sanitize_packages.py
```diff
@@ -0,0 +1,78 @@
...
+            if pkg not in configs.PKG_PY_VERSION_NOT_SUPPORTED[3]:
+                py3_pkgs.append(unquoted_url)
+
+        args = [
+            # ('python', py2_pkgs),
+            ('python3', py3_pkgs),
+        ]
+        for command, packages in args:
+            pip_result = pip_checker.check(
+                [command, '-m', 'pip'], packages, clean=True)
```
liyanhui1228: We will need to use the `fake_pip` in unit test instead of using the real pip.

ylil93: I put a skip test decorator on this test, so it won't run on circle ci
I wrote this test to make sure the urls and pypi packages in the hardcoded whitelists are accurate, but I thought it would be useful to keep this test in this file.

liyanhui1228: OK sounds good, thanks!

PR #188, opened 2018-11-19
add compatibility-lib to whitelists

ylil93:


compatibility_server/configs.py
```diff
@@ -69,10 +69,10 @@ def _format_url(repo_name, setuppy_path=''):
     'tensorboard',
     'tensorflow',
     'gcloud',
-    'compatibility-lib'
 ]
 
-WHITELIST_PKGS = PKG_LIST
+# TODO: add compatibility-lib to PKG_LIST for next release
+WHITELIST_PKGS = PKG_LIST + ['compatibility-lib']
```
liyanhui1228: Why not just adding this package to `PKG_LIST`?

PR #191, opened 2018-11-29
add semver_checker that checks two packages for semver breakage

ylil93:


compatibility_lib/compatibility_lib/semver_checker.py
```diff
@@ -0,0 +1,46 @@
...
+    new_pkg_info = crawler.get_package_info(new_dir)
+
+    unseen = [(old_pkg_info, new_pkg_info)]
+    i = 0
+    while i < len(unseen):
+        old, new = unseen[i]
+        for key in old.keys():
+            if new.get(key) is None:
+                return False
+            if hasattr(old[key], 'get'):
```
liyanhui1228: Could you explain this a little bit?

compatibility_lib/compatibility_lib/testpkgs/added_func/0.1.0/main.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2018 Google LLC
```
liyanhui1228: Probably we need to make more complicated test files to cover all the types we are comparing including modules, classes, functions etc...

compatibility_lib/compatibility_lib/semver_checker.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+# TODO: This needs more sophisticated logic
+# - needs to look at args
+def check(old_dir, new_dir):
+    """checks for semver breakage for two local directories
+
+    Args:
+        old_tag: the tag for an older commit/release
+        new_tag: the tag for a newer commit/release
```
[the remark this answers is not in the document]
ylil93: done

compatibility_lib/compatibility_lib/testpkgs/added_func/0.1.0/main.py
ylil93: Yes I think more test cases are definitely needed, but as to not overwhelm the review process, maybe the complexity of test cases can be scaled up over different PRs?

compatibility_lib/compatibility_lib/semver_checker.py
```diff
@@ -0,0 +1,55 @@
...
+    i = 0
+    while i < len(unseen):
+        old, new = unseen[i]
+        for key in old.keys():
+            if new.get(key) is None:
+                return False
+            if key == 'args':
+                old_args = old[key]
+                new_args = new[key]
+                for arg in old[key]:
```
liyanhui1228: Per discussion, we should make sure the old args and new args are exactly the same, also same in their order.

compatibility_lib/compatibility_lib/semver_checker.py
```diff
@@ -0,0 +1,55 @@
...
+    old_pkg_info = crawler.get_package_info(old_dir)
+    new_pkg_info = crawler.get_package_info(new_dir)
+
+    unseen = [(old_pkg_info, new_pkg_info)]
+    i = 0
+    while i < len(unseen):
+        old, new = unseen[i]
+        for key in old.keys():
+            if new.get(key) is None:
+                return False
```
liyanhui1228: Probably it makes sense to output a message telling why the new version is breaking semantic versioning. Something like "Function foo was removed in new version." This will require checking all the elements in the dictionary instead of return False once it detects a breaking change.

compatibility_lib/compatibility_lib/semver_checker.py
ylil93: done

compatibility_lib/compatibility_lib/semver_checker.py
ylil93: done

PR #197, opened 2018-12-06
Add timestamp for github head badge

liyanhui1228:


badge_server/main.py
```diff
@@ -161,7 +172,7 @@ def one_badge_image():
     if badge_name is None:
         badge_name = package_name
 
-    badge_name = badge_utils._sanitize_badge_name(badge_name)
+    badge_name, is_github = badge_utils._process_github_head_name(badge_name)
```
ylil93: It seems like the purpose of _process_github_head_name() might be confusing?
Maybe tweaking the function name and return value to something like this would be more readable?
```
is_github = badge_utils.is_github_head(badge_name)
if is_github:
    badge_name = badge_utils.GITHUB_HEAD_NAME
```

liyanhui1228: Changed, thanks!

PR #198, opened 2018-12-07
Write github head compatibility status to bigquery

liyanhui1228:


compatibility_lib/compatibility_lib/configs.py
```diff
@@ -75,23 +87,36 @@ def _format_url(repo_name, setuppy_path=''):
...
-    _format_url('googleapis/google-cloud-python', 'kms'): 'gcloud',
-    _format_url('googleapis/google-cloud-python', 'legacy/google-cloud'):
-        'gcloud',
-    _format_url('googleapis/google-cloud-python', 'ndb'): 'gcloud',
-    _format_url('googleapis/google-cloud-python', 'oslogin'): 'gcloud',
-    _format_url('googleapis/google-cloud-python', 'redis'): 'gcloud',
-    _format_url('googleapis/google-cloud-python', 'securitycenter'): 'gcloud',
-    _format_url('googleapis/google-cloud-python', 'tasks'): 'gcloud',
-    _format_url('googleapis/google-cloud-python', 'test_utils'): 'gcloud',
-    _format_url('googleapis/google-cloud-python', 'texttospeech'): 'gcloud',
```
ylil93: Why are these being deleted?

liyanhui1228: These are updated not deleted. Actually wanted to confirm with you is the value in the whitelist dict representing the PyPI names of the packages? And why all of them are `'gcloud'`? I changed them to the corresponding `google-cloud-*` names.

ylil93: Oh sorry my bad - at a quick glance, the way the changes being displayed on github made me think they're being deleted.
Yes the dict values are the PyPI names. I think because the original list only included `gcloud` and not the google-cloud* names, I just assigned the dict values to be gcloud as well.

liyanhui1228: ’gcloud‘ and 'google-cloud-*' are different things actually :) The original list just didn't include all the google cloud client libraries, and I added them in this PR.

Issue #200, opened 2018-12-11
Cache the github head check result based on commit number


liyanhui1228:
- [ ] Show the timestamp for github head version check (#197)
- [ ] Store the github version check result to datastore as a cache based on commit number

PR #201, opened 2018-12-12
Cache the github head version result based on commit number

liyanhui1228:
For #200

### DIFF
--- a/compatibility_lib/compatibility_lib/get_compatibility_data.py
+++ b/compatibility_lib/compatibility_lib/get_compatibility_data.py
@@ -53,25 +53,11 @@ def _result_dict_to_compatibility_result(results, python_version):
     return res_list
 
 
-def _generate_pairs_for_github_head():
-    """Generate pairs for each github head package with the PyPI packages.
-
-    e.g. [(github_pkg, pkg1), (github_pkg, pkg2),...]
-    """
-    pkg_pairs = []
-
-    for gh_pkg in configs.WHITELIST_URLS.keys():
-        gh_pairs = [(gh_pkg, package) for package in configs.PKG_LIST]
-        pkg_pairs.extend(gh_pairs)
-
-    return pkg_pairs
-
-
 def write_to_status_table():
-    """Get the compatibility status for PyPI and github head versions."""
+    """Get the compatibility status for PyPI versions."""
     # Write self compatibility status to BigQuery
     self_res_list = []
-    packages = configs.PKG_LIST + list(configs.WHITELIST_URLS.keys())
+    packages = configs.PKG_LIST
     for py_version in [PY2, PY3]:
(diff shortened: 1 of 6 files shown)


compatibility_lib/compatibility_lib/test_get_compatibility_data.py
```diff
@@ -72,30 +72,6 @@ def mock_init():
             'compatibility_lib.get_compatibility_data.store',
             self.fake_store)
 
-    def test__generate_pairs_for_github_head(self):
```
ylil93: Can you explain why this test is removed?

compatibility_lib/compatibility_lib/get_compatibility_data.py
```diff
@@ -88,14 +74,6 @@ def write_to_status_table():
...
         store.save_compatibility_statuses(res_list)
 
-        # For github head versions
-        pkg_sets = _generate_pairs_for_github_head()
-        results = checker.get_pairwise_compatibility(
-            python_version=py_version,
-            pkg_sets=pkg_sets)
-        res_list = _result_dict_to_compatibility_result(results, py_version)
-        store.save_compatibility_statuses(res_list)
-
```
ylil93: Is the compatibility status no longer saved to the datastore because of the introduced commit number caching?

liyanhui1228: Yep as we will run the check for gh head version to get the most recent check results, we don't need to run it every 24 hours and store into bigquery.

compatibility_lib/compatibility_lib/test_get_compatibility_data.py
liyanhui1228: This is no longer needed, as we don't write to bigquery for gh head versions if we have the cache for commit numbers.

ylil93: Ok - the reason I asked is because I saw you moved the function `_generate_pairs_for_github_head()` to compatibility_lib/compatibility_lib/utils.py

If this function is no longer needed, you should probably remove it from that file as well.

compatibility_lib/compatibility_lib/get_compatibility_data.py
ylil93: Ok cool! Out of curiosity how does the datastore get updated from now on?

liyanhui1228: There is a kokoro job for populating the cache in datastore and update it.

compatibility_lib/compatibility_lib/test_get_compatibility_data.py
liyanhui1228: Done.

PR #202, opened 2018-12-12
semver checker: do not return error msg on new args with default values

ylil93:
- updates package crawler to only include class/function args that don't have default values


compatibility_lib/compatibility_lib/package_crawler_static.py
```diff
@@ -219,9 +219,11 @@ def _get_function_info(functions):
 
 
 def _get_args(node):
-    """returns a list of args"""
+    """returns a list of non-default args, ignoring args with defaults"""
     args = []
-    for arg in node.args:
+    num_args = len(node.args) - len(node.defaults)
```
brianquinlan: Python argument lists can be pretty complex e.g.
def baz(foo, bar=3, *args, key=5): pass

You might want to write a bunch of tests to make sure that you handle all of the edge cases.

ylil93: Oh nice catch - I forgot about that way of passing args.

Quick question about *args and **kwargs though: would a signature like this ever appear?
`foo(a, *args, c): pass`

brianquinlan: I don't know how often that appears in practice. I like to do foo(a, *, b, c, d) because I'm a fan of keyword arguments ;-)

ylil93: i'll add more test cases!

ylil93: done

compatibility_lib/compatibility_lib/semver_checker.py
```diff
@@ -30,7 +29,7 @@ def check(old_dir, new_dir):
         new_dir: directory containing new files
 
     Returns:
-        False if changes breaks semver, True if semver is preserved
+        a list of error strings describing semver breakages
```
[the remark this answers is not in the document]
ylil93: done

compatibility_lib/compatibility_lib/package_crawler_static.py
```diff
@@ -63,9 +63,14 @@ def get_package_info(root_dir):
             module_name: {
                 'classes': {
                     class_name: {
-                        'args': [arg1, arg2, ...],
+                        'args': {
+                            'single_args': [arg1, arg2, ...],
+                            'defaults': {arg1: value1, ...},
+                            'vararg': argv,
+                            'kwarg': kwargs,
```
brianquinlan: Don't you need to know the position of these things?

ylil93: I get the args using ast tooling which returns the args in order.

PR #207, opened 2018-12-17
Add readme for deploying to gke

liyanhui1228:


compatibility_server/configs.py
```diff
@@ -34,28 +34,40 @@ def _format_url(repo_name, setuppy_path=''):
     'google-api-core',
```
brianquinlan: This should go in a different PR.

PR #208, opened 2018-12-18
Handle the case when timestamp str is None

liyanhui1228:


badge_server/utils.py
```diff
@@ -174,8 +174,12 @@ def _calculate_commit_number(package: str) -> Optional[str]:
     return None
 
 
-def _is_github_cache_valid(cache_timestamp_str):
+def _is_github_cache_valid(cache_timestamp_str=None):
```
brianquinlan: Is there a test for this?

liyanhui1228: I'll add one, the test failed in the dashboard building step where it needs the package results existed in bigquery as we added several new packages to the whitelist. It will be fixed by itself when the kokoro job stores the results.

PR #233, opened 2019-01-28
Better catch the error when pip install

liyanhui1228:


compatibility_server/pip_checker.py
```diff
@@ -39,11 +39,10 @@
...
 TIME_OUT = 300  # seconds
 
-# Pattern for pip installation errors not related to the package being
-# installed. See:
-# https://github.com/pypa/pip/blob/3a77bd667cc68935040563e1351604c461ce5333/src/pip/_internal/commands/install.py#L533
-PIP_ENVIRONMENT_ERROR_PATTERN = re.compile(
-    r'not install packages due to an EnvironmentError: (?P<error>.*)')
+# Pattern for pip installation error related to the package being
+# installed.
+PIP_INSTALL_ERROR_PATTERN = re.compile(
```
brianquinlan: If I understand correctly, the idea is that the only time that pip will return a non-zero error code that is package-related is when the package doesn't exist. And you think that is cleaner than the previous approach. Could you document that and maybe point to some source that shows that is true?

compatibility_server/pip_checker.py
```diff
@@ -350,15 +349,14 @@ def _install(self, container: docker.models.containers.Container):
...
-                                  PipCheckResultType.INSTALL_ERROR,
-                                  output)
+            pip_install_error = PIP_INSTALL_ERROR_PATTERN.search(output)
+            if pip_install_error:
+                return PipCheckResult(self._packages,
+                                      PipCheckResultType.INSTALL_ERROR,
+                                      output)
+            raise PipError(error_msg=output,
+                           command=command,
+                           returncode=returncode)
```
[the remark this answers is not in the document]
liyanhui1228: The outer if is checking whether the returncode is 0 or not, and if it is not 0 it will continue to check whether it is related to the pip install itself.

ylil93:
LGTM - please address Brian's comment

compatibility_server/pip_checker.py
liyanhui1228: Yeah I was thinking that if there is a docker timeout error during the pip install process, it will also be treated as an `INSTALL_ERROR`  in the result, which actually isn't. And there might still be some other type of error happened during the installation, probably we could restrict that install error only means the package and version doesn't exist.

brianquinlan: A docker timeout should lead to a `ReadTimeout`, which results in
```raise PipCheckerError(...)```

Are you sure that a non-zero `pip install` return can only be one of:
1. the package/version not existing
2. some sort of internal error

File: compatibility_server/pip_checker.py [hunk 3]
```diff
@@ -372,7 +372,7 @@ def _check(self, container: docker.models.containers.Container):
             command,
             stdout=True,
             stderr=True,
-            raise_on_failure=False)
+            raise_on_failure=True)
```
brianquinlan: I don't think that this will work (see the comment 7 lines up).

liyanhui1228: Yep and probably we could just raise an error for that case, then it will return a 500 and will not be treated as `INSTALL_ERROR`.

liyanhui1228: And I have tested that if I set the docker timeout to a very short time like 1 second, I could reproduce this.

compatibility_server/pip_checker.py
```diff
@@ -350,12 +350,19 @@ def _install(self, container: docker.models.containers.Container):
...
             if environment_error:
                 raise PipError(error_msg=environment_error.group('error'),
                                command=command,
                                returncode=returncode)
 
+            # Checking for error caused by sigkill (128+9)
+            if returncode >= 128 and returncode <= 137:
+                raise PipCheckerError(
+                    error_msg="The docker container timed out before executing"
+                              " pip command. Error msg: {}".format(output))
```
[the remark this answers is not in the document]

compatibility_server/pip_checker.py
```diff
@@ -350,12 +350,19 @@ def _install(self, container: docker.models.containers.Container):
...
             environment_error = PIP_ENVIRONMENT_ERROR_PATTERN.search(output)
             if environment_error:
                 raise PipError(error_msg=environment_error.group('error'),
                                command=command,
                                returncode=returncode)
 
+            # Checking for error caused by sigkill (128+9)
+            if returncode >= 128 and returncode <= 137:
+                raise PipCheckerError(
+                    error_msg="The docker container timed out before executing"
```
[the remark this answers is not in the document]

compatibility_server/pip_checker.py
```diff
@@ -350,12 +350,19 @@ def _install(self, container: docker.models.containers.Container):
...
         if returncode:
+            # Checking for environment error
             environment_error = PIP_ENVIRONMENT_ERROR_PATTERN.search(output)
             if environment_error:
                 raise PipError(error_msg=environment_error.group('error'),
                                command=command,
                                returncode=returncode)
 
+            # Checking for error caused by sigkill (128+9)
+            if returncode >= 128 and returncode <= 137:
```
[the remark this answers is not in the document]

compatibility_server/pip_checker.py
```diff
@@ -350,12 +350,19 @@ def _install(self, container: docker.models.containers.Container):
...
             raise_on_failure=False)
         if returncode:
+            # Checking for environment error
             environment_error = PIP_ENVIRONMENT_ERROR_PATTERN.search(output)
             if environment_error:
                 raise PipError(error_msg=environment_error.group('error'),
                                command=command,
                                returncode=returncode)
 
+            # Checking for error caused by sigkill (128+9)
```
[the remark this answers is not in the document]

brianquinlan:
Looks good but shouldn't there be a test for this?

compatibility_server/test_pip_checker.py
```diff
@@ -95,8 +94,8 @@ def exec_run(self, cmd, stdout=True, stderr=True):
         duration = current_time - self.start_time
 
         if duration > pip_checker.TIME_OUT:
-            raise docker.errors.APIError(message="time out",
-                                         explanation="Request time out.")
```
liyanhui1228: This wasn't mocking the docker container timeout behavior correctly before, which should actually return a 137 code instead of raising a `docker.errors.APIError`. Changing this to match the behavior and the test in line 123 is for testing the timeout error.

compatibility_server/pip_checker.py
```diff
@@ -313,7 +313,18 @@ def _run_command(
...
 
-        if returncode and raise_on_failure:
+        # Checking for cases where the command was killed by a signal.
+        # If a process was killed by a signal, then it's exit code will be
+        # 128 + <signal number>.
+        # If a docker container exits with a running command then it will be
+        # killed with SIGKILL => 128 + 9 = 137
+        if returncode > 128 and returncode <= 137:
+            raise PipCheckerError(
+                error_msg="The command {} was killed by the signal {}."
```
brianquinlan:
```suggestion
                error_msg="The command {} was killed by the signal {}. "
```

brianquinlan:
```suggestion
                error_msg="The command {} was killed by signal {}."
```

compatibility_server/pip_checker.py
```diff
@@ -313,7 +313,18 @@ def _run_command(
...
-        if returncode and raise_on_failure:
+        # Checking for cases where the command was killed by a signal.
+        # If a process was killed by a signal, then it's exit code will be
+        # 128 + <signal number>.
+        # If a docker container exits with a running command then it will be
+        # killed with SIGKILL => 128 + 9 = 137
+        if returncode > 128 and returncode <= 137:
+            raise PipCheckerError(
+                error_msg="The command {} was killed by signal {}. "
+                          "This likely means that the Docker container timed"
```
brianquinlan:
```suggestion
                          "This likely means that the Docker container timed "
```

liyanhui1228: done

dashboard/dashboard_builder.py
```diff
@@ -306,7 +306,7 @@ def get_result(self,
                         'self': False,
                     }
                 )
-                if status_type is 'self-success':
```
brianquinlan: Nice catch!

PR #235, opened 2019-01-30
Add cloud build configs for generating dashboard

liyanhui1228:


liyanhui1228:
> Maybe write some docs explaining the approach?

Done.

Issue #237, opened 2019-02-02
Install apache-beam[gcp] github head in py3 will fail


liyanhui1228:
```
$ pip3 install git+git://github.com/apache/beam.git#subdirectory=sdks/python
Collecting git+git://github.com/apache/beam.git#subdirectory=sdks/python
  Cloning git://github.com/apache/beam.git to /tmp/pip-req-build-ovp76cpa
apache-beam requires Python '>=2.7,<3.0' but the running Python is 3.6.3
```

PR #238, opened 2019-02-02
Filter out the packages not supported

liyanhui1228:
For #237


ylil93:
LGTM, but please fix failures

PR #239, opened 2019-02-06
Add sql file for creating tables

liyanhui1228:


sql_schema/create_table.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE self_compatibility_status (install_name VARCHAR(200), status VARCHAR(20), py_version VARCHAR(5), timestamp TIMESTAMP, details VARCHAR(5000), PRIMARY KEY (install_name, py_version));
```
brianquinlan: Is there a text type that you can use instead of VARCHAR(5000)? I think that you only need VARCHAR to support queries). Why is py_version 5 characters?

sql_schema/create_table.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE self_compatibility_status (install_name VARCHAR(200), status VARCHAR(20), py_version VARCHAR(1), timestamp TIMESTAMP, details TEXT, PRIMARY KEY (install_name, py_version));
```
brianquinlan: It seems like, in the production table, the timestamp is auto-updating?

PR #240, opened 2019-02-08
Generate package pairs to check

liyanhui1228:
This is separating out from #236


liyanhui1228:
Tests WIP, will push in a later commit.

PR #241, opened 2019-02-08
Build python image with pre-stored cloud clientlibs

liyanhui1228:
When `pip install git+git://github.com/googleapis/google-cloud-python.git#subdirectory=[NAME]`, it needs to clone the entire git repository of google-cloud-python instead of only the subdirectory we want to check, which makes the check very slow for checking github head version cloud client libs.

This PR added a build step for building our own python2.7 and python3.6 images, with the github head version package stored in those docker images. And when checking the github head version for cloud client libs, it will do `pip install -e google-cloud-python-master/[NAME]`, which install the package from source.

We will need to set up a kokoro job to build the py27 and py36 images every 5 minutes, to make sure we have the latest version of the github package to check.


cloud_build/py27_docker/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM python:2.7
+
+RUN wget https://github.com/googleapis/google-cloud-python/archive/master.zip
```
wangxf123456: Maybe we can use a chained RUN in Dockerfile?

Reason: https://stackoverflow.com/questions/39223249/multiple-run-vs-single-chained-run-in-dockerfile-what-is-better
Example: https://github.com/GoogleCloudPlatform/cloud-sdk-docker/blob/master/Dockerfile

liyanhui1228: Done.

PR #242, opened 2019-02-11
Add script for cleaning up old python images

liyanhui1228:
This is for cleaning up the image created by #241

### DIFF
--- a/scripts/clean_up_images.py
+++ b/scripts/clean_up_images.py
@@ -1,7 +1,7 @@
 import logging
 import subprocess
 
-LIST_IMAGES_COMMAND = 'gcloud container images list-tags'
+LIST_IMAGES_COMMAND_PREFIX = 'gcloud container images list-tags'
 DELETE_IMAGES_COMMAND_PREFIX = 'gcloud container images delete'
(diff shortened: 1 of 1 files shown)


scripts/clean_up_images.py
```diff
@@ -0,0 +1,36 @@
...
+    digests_list = digests.decode('utf-8').strip('\n').split('\n')
+    num_of_digests = len(digests_list)
+
+    if num_of_digests > ALLOWED_SIZE:
+        to_delete = digests_list[ALLOWED_SIZE:]
+        print(to_delete)
+        for digest in to_delete:
+            command = DELETE_IMAGES_COMMAND_PREFIX + \
+                ' gcr.io/python-compatibility-tools/{}@{}'.format(image_name, digest) \
+                + ' --force-delete-tags'
```
ylil93: It might be a good idea to ask for confirmation after displaying the images that will be deleted before executing the command.

liyanhui1228: Thanks for reviewing! I'm going to close this PR as #243 is checked in.

PR #243, opened 2019-02-12
Shallow clone the github head version for client libs

liyanhui1228:
An alternative for #241

### DIFF
--- a/compatibility_server/pip_checker.py
+++ b/compatibility_server/pip_checker.py
@@ -38,6 +38,9 @@ import docker
 PYPI_URL = 'https://pypi.org/pypi/'
 CONTAINER_WITH_PKG = "checker"
 TIME_OUT = 300  # seconds
+GITHUB_CLIENTLIBS_PREFIX = 'git+git://github.com/googleapis/' \
+                           'google-cloud-python.git#subdirectory='
+CLIENTLIBS_GITHUB_URL = 'https://github.com/googleapis/google-cloud-python.git'
 
 # Pattern for pip installation errors not related to the package being
 # installed. See:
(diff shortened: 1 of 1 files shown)


compatibility_server/pip_checker.py
```diff
@@ -38,6 +38,9 @@
 PYPI_URL = 'https://pypi.org/pypi/'
 CONTAINER_WITH_PKG = "checker"
 TIME_OUT = 300  # seconds
+GITHUB_CLIENTLIBS_PREFIX = 'git+git://github.com/googleapis/' \
```
brianquinlan: Is this really called client libs? I thought that was https://github.com/googleapis/google-api-python-client

compatibility_server/pip_checker.py
```diff
@@ -353,24 +356,44 @@ def _build_command(self, subcommands: List[str]):
...
-                raise PipError(error_msg=environment_error.group('error'),
-                               command=command,
-                               returncode=returncode)
-
-            return PipCheckResult(self._packages,
-                                  PipCheckResultType.INSTALL_ERROR,
-                                  output)
+        # If there is cloud client libraries, install it from source.
+        # Else install it from PyPI.
+        for pkg in self._packages:
```
brianquinlan: I'd structure this a bit differently to make it more readable.

```python

def _clone_repo(self, container, repo):
 """<docstring>"""
  _, directory = self._run_command(
    container,
    "echo $(mktemp -d)"
    stderr=False,
    stdout=True,
    raise_on_failure=True)

  shallow_clone_command = ['git', 'clone', '--depth', '1', repo, directory]
  self._run_command(
    container,
    shallow_clone_command,
    stdout=False,
    stderr=True,
    raise_on_failure=True)
  return directory

def _install(self, container: docker.models.containers.Container):
if any(GITHUB_CLIENTLIBS_PREFIX in pkg for pkg in self._packages):
  # The https://github.com/googleapis/google-cloud-python repository is large,
  # has a large amount of history and contains many independent packages.
  # Installing packages from that repository is quite slow e.g.
  # $ time pip install git+git://...
  # <result>
  # As an optimization, if any provided package is installed from this repository,
  # it is first cloned (using --depth=1 to avoid unnecessary history) and then
  # installed from local disk.
  client_repo_directory = self._clone_repo(CLIENTLIBS_GITHUB_URL)

install_names = []
for pkg in self._packages:
  if GITHUB_CLIENTLIBS_PREFIX in pkg:
    install_subdirectory = pkg.split(GITHUB_CLIENTLIBS_PREFIX)[1]
    install_names.append(os.path.join(client_repo_directory, install_subdirectory))
  else:
    install_names.append(pkg)
command = self._build_command(['install', '-U'] + install_names)
# Rest is the same.
```

compatibility_server/pip_checker.py
```diff
@@ -353,24 +356,44 @@ def _build_command(self, subcommands: List[str]):
...
+            if GITHUB_CLIENTLIBS_PREFIX in pkg:
+                # Shallow clone the repository
+                shallow_clone_command = ['git', 'clone', '--depth', '1',
+                                         CLIENTLIBS_GITHUB_URL]
+                self._run_command(
```
brianquinlan: I don't like that we can run this multiple times and every time after the first will fail. I also don't like that we don't check for failures.

compatibility_server/pip_checker.py
liyanhui1228: It's https://github.com/googleapis/google-cloud-python actually, I just call it clientlibs for short. That one is the discovery based APIs.

compatibility_server/pip_checker.py

PR #246, opened 2019-02-14
Stop building gh-pages branch

liyanhui1228:


liyanhui1228:
> Are you planning on deleting the gh-pages branch? I think that would be a good idea.

Yep.

PR #249, opened 2019-02-16
update configs file

ylil93:
- add comment explaining WHITELIST_URL
- add test to make sure config files match


compatibility_server/configs.py
```diff
@@ -1,187 +0,0 @@
-# Copyright 2018 Google LLC
```
liyanhui1228: The compatibility server is deployed using docker, and it won't be able to find the symlinks without uploading the ../compatibility_lib/ stuff with the same directory structure unfortunately :(

ylil93: That's a good point - what if compatibility_lib/compatibility_lib/configs.py is the symlink file instead?

liyanhui1228: That's the same, releasing the compatibility_lib to PyPI won't include the files in compatibility_server.

ylil93: hmm I guess they're pretty isolated from each other

Maybe I can add a test to catch any differences instead.

liyanhui1228: SGTM.

ylil93: done

PR #250, opened 2019-02-19
fix docker issues

ylil93:
- take export_metrics flag via environment variable
- update requirements.txt with hidden dependencies from opencensus

### DIFF
--- a/compatibility_server/compatibility_checker_server.py
+++ b/compatibility_server/compatibility_checker_server.py
@@ -197,7 +197,8 @@ def main():
         help='port to which the server should bind')
     export_metrics = os.environ.get('EXPORT_METRICS') is not None
 
-    argsdict = vars(parser.parse_args())
+    args = parser.parse_args()
+    argsdict = vars(args)
     argsdict['export_metrics'] = export_metrics
     logging.info('Running server with:\n%s', pprint.pformat(argsdict))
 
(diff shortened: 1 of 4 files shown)


compatibility_server/compatibility_checker_server.py
```diff
@@ -194,14 +195,14 @@ def log_message(self, format, *args):
         type=int,
         default=8888,
         help='port to which the server should bind')
-    parser.add_argument(
-        '--export_metrics',
-        action='store_true',
-        help='whether or not metrics should be exported to stackdriver')
+    export_metrics = os.environ.get('EXPORT_METRICS') is not None
```
liyanhui1228: Is `is not None` necessary here?

compatibility_server/requirements.txt
```diff
@@ -3,3 +3,5 @@ docker
 flask
 requests
 opencensus==0.2.0
+grpcio==1.18.0
```
liyanhui1228: Sort the packages in order?

compatibility_server/compatibility_checker_server.py
ylil93: Since `export_metrics` will be logged, I think it makes sense to have the flag be either `True` or `False` instead of something arbitrary for log readability.

compatibility_server/requirements.txt
ylil93: Done

PR #251, opened 2019-02-21
Add timeout for compatibility_checker

liyanhui1228:


ylil93:
So what would happen from the user's point of view when the compatibility checker times out?

liyanhui1228:
> So what would happen from the user's point of view when the compatibility checker times out?

It will retry by itself.

PR #254, opened 2019-02-21
Make dashboard builder able to use cloud sql

liyanhui1228:


dashboard/dashboard_builder.py
```diff
@@ -356,43 +367,60 @@ def main():
...
+    cloud_sql_proxy_path = './cloud_sql_proxy'
+
+    try:
+        # Run cloud_sql_proxy
+        process = popen_spawn.PopenSpawn([cloud_sql_proxy_path, instance_flag])
+        process.expect('Ready for new connection', timeout=5)
+
+        packages = [
+            package.Package(install_name) for install_name in args.packages]
+        logging.info("Getting self compatibility results...")
```
wangxf123456: Be consistent when using quotation marks.

scripts/update_dashboard.sh
```diff
@@ -18,6 +18,8 @@ set -ev
 # Build dashboard
 function build_dashboard {
     if [ -n "$GOOGLE_APPLICATION_CREDENTIALS" ]; then
+      wget https://dl.google.com/cloudsql/cloud_sql_proxy.linux.amd64 -O cloud_sql_proxy
+      chmod 777 cloud_sql_proxy
```
wangxf123456: chmod +x cloud_sql_proxy

dashboard/dashboard_builder.py
```diff
@@ -356,43 +367,60 @@ def main():
...
-        'dashboard/main-template.html')
-
-    main_path = os.path.dirname(os.path.abspath(__file__)) + '/index.html'
-    with open(main_path, 'wt') as f:
-        f.write(main_html)
+    instance_flag = '-instances={}=tcp:{}'.format(
+        INSTANCE_CONNECTION_NAME, PORT)
+    cloud_sql_proxy_path = './cloud_sql_proxy'
+
+    try:
```
ylil93: It seems like this main() function is quite long and is significantly over the 40 lines recommended by the internal style guide. I suggest breaking up this chunk of code into smaller functions unless doing so would harm the structure of the program (which seems doable to me).

scripts/update_dashboard.sh
liyanhui1228: Done.

dashboard/dashboard_builder.py
liyanhui1228: Done.

dashboard/dashboard_builder.py
liyanhui1228: I'll fix the style in a later PR. Can we merge it for now as the dashboard builder is failing because of now able to access mysql...

liyanhui1228: I added a TODO in there as a reminder.

ylil93: SGTM

PR #260, opened 2019-02-23
Make badge server connect to cloud sql

liyanhui1228:


badge_server/utils.py
```diff
@@ -33,8 +33,12 @@
 from compatibility_lib import deprecated_dep_finder
 
 # Initializations
+DB_CONNECTION_NAME = 'python-compatibility-tools:us-central1:' \
+                     'compatibility-data'
+UNIX_SOCKET = '/cloudsql/{}'.format(DB_CONNECTION_NAME)
+
 checker = compatibility_checker.CompatibilityChecker()
-store = compatibility_store.CompatibilityStore()
+store = compatibility_store.CompatibilityStore(mysql_unix_socket=UNIX_SOCKET)
```
ylil93: 1. Confirming what was said in chat, for local testing, you can instantiate CompatibilityStore() as
`store = compatibility_store.CompatibilityStore()`

2. Does this line imply that references to CompatibilityStore on the badge server should be done like this?
```
import utils

store = utils.store
```

liyanhui1228: 1. Yes.
2. It's flexible to change according to the use cases. Currently there is no need to access store object directly from main.py

PR #261, opened 2019-02-23
A few clean-ups

liyanhui1228:


compatibility_lib/compatibility_lib/compatibility_store.py
```diff
@@ -130,6 +130,7 @@ def __init__(self,
                  mysql_user=None,
                  mysql_password=None,
                  mysql_host=None,
+                 mysql_port=None,
```
ylil93: Nice - I was going to do a PR for this as well!

Is there a reason the default value is set to `None` rather than `3306`?
It seems redundant to have this appear later on:
```
if mysql_port is None:
    mysql_port = 3306
```

liyanhui1228: Done

PR #266, opened 2019-02-27
Remove obsolete files to build raw VM docker instances

brianquinlan:


liyanhui1228:
We could ignore the test failures for this PR. As this one is not merging from a forked branch, the build_dashboard will run on this PR, which is sure to fail. (There is some cloud_sql_proxy issue in there, but as we are going to remove the build dashboard workflow from github soon, I didn't fix it).

PR #273, opened 2019-03-14
Badge server: prepare for PRD changes

ylil93:
Specifically, the left hand text, right hand text, and badge coloring need to change. I thought this would be relatively straightforward, but since all three attributes are bundled together in a `CompatibilityResult` object, the code around pip_check in the compatibility server will have to change and possibly the logic in creating `CompatibilityResult` objects in compatibility store/cloudsql would have to change as well.

### DIFF
--- a/badge_server/main.py
+++ b/badge_server/main.py
@@ -77,13 +77,16 @@ def _get_self_compatibility_dict(package_name: str) -> dict:
 
 
 def _get_other_package_from_set(name: str,
-                                package_set: frozenset) -> package.Package:
-    """Returns the other package.
+                                package_set: frozenset[package.Package]
+                                ) -> package.Package:
+    """Returns the package that does *not* have the given name.
 
     Args:
-        name: a string, the given package name
+        name: The name of the package not to return.
+        package_set: A set of two unsorted packages, one has the given name.
+
     Returns:
-        The Package object that doesn't correspond to the give package name
+        The Package object that doesn't correspond to the give package name.
     """
     first, second = package_set
     if first.install_name == name:
--- a/badge_server/main.py
+++ b/badge_server/main.py
@@ -275,7 +275,7 @@ def _format_badge_name(package_name, badge_name, commit_number):
     if 'github.com' in package_name:
         return 'compatibility check (master)'
     else:
-        return 'compatibility check (pypi)'
+        return 'compatibility check (PyPI)'
 
 
 @app.route('/one_badge_image')
--- a/badge_server/main.py
+++ b/badge_server/main.py
@@ -35,6 +35,7 @@ Example Usage:
     http://0.0.0.0:8080/one_badge_target?package=tensorflow
 """
 
+import enum
 import flask
 import pybadges
 
--- a/badge_server/main.py
+++ b/badge_server/main.py
@@ -84,7 +84,8 @@ def _get_other_package_from_set(name: str,
 
     Args:
         name: The name of the package not to return.
-        package_set: A set of two unsorted packages, one has the given name.
+        package_set: A set of two unsorted packages, one of which has the
+            given name.
 
     Returns:
         The Package object that doesn't correspond to the give package name.
--- a/badge_server/main.py
+++ b/badge_server/main.py
@@ -47,6 +47,41 @@ from typing import FrozenSet
 app = flask.Flask(__name__)
 
 
+@enum.unique
+class BadgeStatus(enum.Enum):
+    """Represents a package's badge status.
+
+    The status is based on the results of running 'pip install' and
+    'pip check' on the compatibility server.
+
+    UNKNOWN_PACKAGE: package not in whitelist
+    INTERNAL_ERROR: unexpected internal error
+    MISSING_DATA: missing package data from package store
+    SELF_INCOMPATIBLE: pip error when installing self
+    INCOMPATIBLE: pip error when installed with another package
+    OUTDATED_DEPENDENCY: package has an outdated dependency
+    SUCCESS: No issues
+    """
+    UNKNOWN_PACKAGE = 'UNKNOWN_PACKAGE'
+    INTERNAL_ERROR = 'INTERNAL_ERROR'
+    MISSING_DATA = 'MISSING_DATA'
+    SELF_INCOMPATIBLE = 'SELF_INCOMPATIBLE'
+    INCOMPATIBLE = 'INCOMPATIBLE'
+    OUTDATED_DEPENDENCY = 'OUTDATED_DEPENDENCY'
+    SUCCESS = 'SUCCESS'
+
+
+PACKAGE_STATUS_TO_COLOR = {
+    BadgeStatus.UNKNOWN_PACKAGE: 'lightgrey',
+    BadgeStatus.INTERNAL_ERROR: 'lightgrey',
+    BadgeStatus.MISSING_DATA: 'lightgrey',
+    BadgeStatus.SELF_INCOMPATIBLE: 'red',
+    BadgeStatus.INCOMPATIBLE: 'red',
+    BadgeStatus.OUTDATED_DEPENDENCY: 'orange',
+    BadgeStatus.SUCCESS: 'green',
+}
+
+
 def _get_self_compatibility_dict(package_name: str) -> dict:
     """Returns a dict containing self compatibility status and details.
 
--- a/badge_server/main.py
+++ b/badge_server/main.py
@@ -42,6 +42,7 @@ import utils as badge_utils
 from compatibility_lib import utils as compat_utils
 from compatibility_lib import configs
 from compatibility_lib import package
+from typing import FrozenSet
 
 app = flask.Flask(__name__)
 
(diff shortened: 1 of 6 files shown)


compatibility_lib/compatibility_lib/configs.py
```diff
@@ -12,7 +12,41 @@
...
+
+    UNKNOWN_PACKAGE: package not in whitelist
+    INTERNAL_ERROR: unexpected internal error
+    MISSING_DATA: missing package data from package store
+    SELF_INCOMPATIBLE: pip error when installing self
+    INCOMPATIBLE: pip error when installed with another package
+    OUTDATED_DEPENDENCY: package has an outdated dependency
+    SUCCESS: No issues
+    """
+    UNKNOWN_PACKAGE = 'lightgrey'
```
brianquinlan: Using the color as the value is odd. I'd have another data structure like:

PACKAGE_STATUS_TO_COLOR = {
  PackageStatus.UNKNOWN_PACKAGE: '''
}

badge_server/main.py
```diff
@@ -76,6 +76,21 @@ def _get_self_compatibility_dict(package_name: str) -> dict:
     return result_dict
 
 
+def _get_other_package_from_set(name: str,
+                                package_set: frozenset) -> package.Package:
+    """Returns the other package.
+
+    Args:
+        name: a string, the given package name
+    Returns:
```
brianquinlan: Blank line before `Returns:`

badge_server/main.py
```diff
@@ -76,6 +76,21 @@ def _get_self_compatibility_dict(package_name: str) -> dict:
     return result_dict
 
 
+def _get_other_package_from_set(name: str,
+                                package_set: frozenset) -> package.Package:
+    """Returns the other package.
```
```suggestion
    """Returns the package that does *not* have the given name.
```

badge_server/main.py
```diff
@@ -76,6 +76,21 @@ def _get_self_compatibility_dict(package_name: str) -> dict:
     return result_dict
 
 
+def _get_other_package_from_set(name: str,
+                                package_set: frozenset) -> package.Package:
```
```suggestion
                                package_set: frozenset[package.Package]) -> package.Package:
```

badge_server/main.py
```diff
@@ -76,6 +76,21 @@ def _get_self_compatibility_dict(package_name: str) -> dict:
...
+    """Returns the other package.
+
+    Args:
+        name: a string, the given package name
```
brianquinlan: You don't have to duplicate the type annotation.
```suggestion
        name: The name of the package not to return.
```

brianquinlan: You must document all arguments.

compatibility_server/configs.py
```diff
@@ -12,7 +12,41 @@
...
+The reason for this set up is that these modules need to be isolated from
+each other, but there also needs to be consistency in the objects and data
+in this file since they exist in the same workflow.
+"""
+
+
+import enum
+
+
+class PackageStatus(enum.Enum):
```
brianquinlan: I still don't understand why this is here?

compatibility_server/pip_checker.py
```diff
@@ -87,6 +87,7 @@ def __init__(self,
         self.command_string = command_string
 
 
+# TODO: deprecate the following:
```
brianquinlan: Why?

badge_server/main.py

badge_server/main.py

badge_server/main.py

badge_server/main.py

compatibility_lib/compatibility_lib/configs.py
ylil93: done

compatibility_server/pip_checker.py
ylil93: This will be replaced by the `PackageStatus` enum introduced in this PR in a future PR.

compatibility_server/configs.py
ylil93: This will replace `PipCheckResultType` in compatibility_server and `Status` in compatibility_store. They are equivalent to each other in the current workflow and need to be expanded upon to support the changes listed in the PRD. I chose to create `PackageStatus` in this file since this file specifically contains items that need to be consistent among compatibility_store and compatibility_lib.

badge_server/main.py
```diff
@@ -76,6 +77,24 @@ def _get_self_compatibility_dict(package_name: str) -> dict:
...
 
 
+def _get_other_package_from_set(name: str,
+                                package_set: FrozenSet[package.Package]
+                                ) -> package.Package:
+    """Returns the package that does *not* have the given name.
+
+    Args:
+        name: The name of the package not to return.
+        package_set: A set of two unsorted packages, one has the given name.
```
```suggestion
        package_set: A set of two unsorted packages, one of which has the given name.
```

compatibility_server/configs.py
brianquinlan: I think that the backend should be, to the extent reasonable, independent of the frontend.

compatibility_lib/compatibility_lib/configs.py
```diff
@@ -12,7 +12,52 @@
...
+    UNKNOWN_PACKAGE = 'UNKNOWN_PACKAGE'
+    INTERNAL_ERROR = 'INTERNAL_ERROR'
+    MISSING_DATA = 'MISSING_DATA'
+    SELF_INCOMPATIBLE = 'SELF_INCOMPATIBLE'
+    INCOMPATIBLE = 'INCOMPATIBLE'
+    OUTDATED_DEPENDENCY = 'OUTDATED_DEPENDENCY'
+    SUCCESS = 'SUCCESS'
+
+
+PACKAGE_STATUS_TO_COLOR = {
```
brianquinlan: Aren't colors a front-end concern that have nothing to do with this library?

compatibility_lib/compatibility_lib/configs.py
```diff
@@ -12,7 +12,52 @@
...
+The reason for this set up is that these modules need to be isolated from
+each other, but there also needs to be consistency in the objects and data
+in this file since they exist in the same workflow.
+"""
+
+
+import enum
+
+
+class PackageStatus(enum.Enum):
```
brianquinlan: Does this belong here or in main.py?

compatibility_server/configs.py
ylil93: As I understand it, the backend runs a pip check on a given package(s) and that is where the status is generated. The front end uses that status to serve a badge. I don't see how the backend status can be different from the front end status. In other words, because the check is done by the backend, if the possible statuses generated by the backend are [UNKNOWN, SUCCESS], the front end won't be able to serve a badge status of INCOMPATIBLE.

Before, compatibility_server and compatibility_lib independently defined enums to capture the possible statuses, but this is not good practice since if one thing changes in the backend, the frontend would break without the same change being applied.

compatibility_lib/compatibility_lib/configs.py
ylil93: Oh good point. I committed this change based on your [previous comment](https://github.com/GoogleCloudPlatform/cloud-opensource-python/pull/273#discussion_r266176591) but I should have caught that too.

I'll move this dict somewhere else.

compatibility_server/configs.py
brianquinlan: The backend has only 3 statuses: success, install_error, check_warning. It doesn't figure out if the dependencies are outdated, it doesn't care if it is checking one or more packages. And it doesn't know anything about storage (i.e. MISSING_DATA).

ylil93: But we're getting rid of install_error and check_warning and adding INTERNAL_ERROR, SELF_INCOMPATIBLE, and INCOMPATIBLE?

brianquinlan: But why are does it make sense to make this change? Most of this enum doesn't apply to the compatibility server but the result that it currently returns makes sense based on what it actually does.

badge_server/main.py

compatibility_server/configs.py
ylil93: I thought of a compromise - take a look at the latest commit and let me know what you think.

compatibility_lib/compatibility_lib/compatibility_store.py
```diff
@@ -33,6 +33,42 @@
...
+    UNKNOWN_PACKAGE = 'UNKNOWN_PACKAGE'
+    INTERNAL_ERROR = 'INTERNAL_ERROR'
+    MISSING_DATA = 'MISSING_DATA'
+    SELF_INCOMPATIBLE = 'SELF_INCOMPATIBLE'
+    INCOMPATIBLE = 'INCOMPATIBLE'
+    OUTDATED_DEPENDENCY = 'OUTDATED_DEPENDENCY'
+    SUCCESS = 'SUCCESS'
+
+
+PACKAGE_STATUS_TO_COLOR = {
```
brianquinlan: Shouldn't these mapping and the enum live in main.py?

compatibility_lib/compatibility_lib/compatibility_store.py
```diff
@@ -33,6 +33,42 @@
...
+    PackageStatus.INTERNAL_ERROR: 'lightgrey',
+    PackageStatus.MISSING_DATA: 'lightgrey',
+    PackageStatus.SELF_INCOMPATIBLE: 'red',
+    PackageStatus.INCOMPATIBLE: 'red',
+    PackageStatus.OUTDATED_DEPENDENCY: 'orange',
+    PackageStatus.SUCCESS: 'green',
+}
+
+
+# TODO: deprecate the following:
```
brianquinlan: Still?

PR #276, opened 2019-03-23
use BadgeStatus instead of strings for status

ylil93:


badge_server/main.py
```diff
@@ -280,22 +286,30 @@ def _get_badge_status(
...
+    if all(status == BadgeStatus.SUCCESS for status in statuses):
+        return BadgeStatus.SUCCESS
 
-    if all(status == 'SUCCESS' for status in statuses):
-        return 'SUCCESS'
-    elif 'UNKNOWN' in statuses:
-        return 'UNKNOWN'
-    return 'CHECK_WARNING'
+    # This works because the enums are stored in an ordered dict.
+    for status in BadgeStatus.__members__.values():
```
brianquinlan: Maybe make a new function: `get_highest_priority_status`?

badge_server/main.py
```diff
@@ -115,18 +116,20 @@ def _get_self_compatibility_dict(package_name: str) -> dict:
...
     compatibility_results = badge_utils.store.get_self_compatibility(pkg)
     result_dict = badge_utils._build_default_result(
-        status='SUCCESS',
+        status=BadgeStatus.SUCCESS,
         details='The package does not support this version of python.')
     for res in compatibility_results:
         pyver = badge_utils.PY_VER_MAPPING[res.python_major_version]
-        result_dict[pyver]['status'] = res.status.value
+        badge_status = PACKAGE_STATUS_TO_BADGE_STATUS.get(
+            res.status) or BadgeStatus.SELF_INCOMPATIBLE
```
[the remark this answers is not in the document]
ylil93: `res.status` would be one of [these](https://github.com/GoogleCloudPlatform/cloud-opensource-python/blob/5bfba65520f3e4d9dd4f4230337d786f50ef007b/compatibility_lib/compatibility_lib/compatibility_store.py#L36), so we want to convert it to a `BadgeStatus`. Other statuses have a direct equivalent, but in the case of `INSTALL_ERROR` the `BadgeStatus` equivalent would be either `SELF_INCOMPATIBLE` or `INCOMPATIBLE` depending on where it's called. Since this is being called from `_get_self_compatibility_dict()` that's why the default is `SELF_INCOMPATIBLE` here.

badge_server/main.py
```diff
@@ -211,11 +214,13 @@ def _get_pair_compatibility_dict(package_name: str) -> dict:
...
             # the `other_package` will be incompatible with all other packages.
             self_compat_res = _get_self_compatibility_dict(other_package_name)
-            if self_compat_res[pyver]['status'] != 'SUCCESS':
+            if self_compat_res[pyver]['status'] != BadgeStatus.SUCCESS:
                 continue
 
             details = res.details or badge_utils.EMPTY_DETAILS
-            result_dict[pyver]['status'] = res.status.value
+            badge_status = PACKAGE_STATUS_TO_BADGE_STATUS.get(
+                res.status) or BadgeStatus.PAIR_INCOMPATIBLE
```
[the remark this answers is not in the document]
ylil93: Since this is being called from `_get_pair_compatibility_dict()` that's why the default is `PAIR_INCOMPATIBLE` here.

badge_server/main.py
```diff
@@ -280,22 +286,30 @@ def _get_badge_status(
...
-    See badge_utils.STATUS_COLOR_MAPPING.
-    """
-    dep_status = dependency_res['status']
-    dep_status = 'SUCCESS' if dep_status == 'UP_TO_DATE' else dep_status
+    Args:
+        self_compat_res: a dict containing a package's self compatibility
+            status for py2 and py3. See _get_self_compatibility_dict().
+        google_compat_res: a dict containing a package's pair compatibility
+            status for py2 and py3. See _get_pair_compatibility_dict().
+        dependency_res: a dict containing a package's dependency status
```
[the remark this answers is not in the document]
ylil93: That was a typo - nice catch!

badge_server/main.py
```diff
@@ -280,22 +286,30 @@ def _get_badge_status(
     The badge status will determine the right hand text and the color of
```
brianquinlan: I don't think that the return type is a string anymore

ylil93: Can I actually do this in a follow up PR?

badge_server/main.py
ylil93: Can I actually do this in a follow up PR?

badge_server/main.py
ylil93: Done!

badge_server/main.py
brianquinlan: Sure.

PR #277, opened 2019-03-27
define get_highest_status() under BadgeStatus

ylil93:
- refactor code to use this method

### DIFF
--- a/badge_server/main.py
+++ b/badge_server/main.py
@@ -92,7 +92,7 @@ class BadgeStatus(enum.Enum):
             contained INTERNAL_ERROR and OUTDATED_DEPENDENCY.
 
         Raises:
-            ValueError if no BadgeStatus exists. For example, if the length of
+            ValueError: If no BadgeStatus exists. For example, if the length of
                 `statuses` is 0, an error is raised.
         """
         for status in cls.__members__.values():
--- a/badge_server/main.py
+++ b/badge_server/main.py
@@ -85,9 +85,11 @@ class BadgeStatus(enum.Enum):
 
         Args:
             statuses: A list of BadgeStatuses.
+
         Raises:
             ValueError if no BadgeStatus exists. For example, if the length of
                 `statuses` is 0, an error is raised.
+
         Returns:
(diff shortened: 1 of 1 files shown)


badge_server/main.py
```diff
@@ -76,6 +77,23 @@ class BadgeStatus(enum.Enum):
...
     SUCCESS = 'SUCCESS'
 
+    @classmethod
+    def get_highest_status(cls, statuses: Iterable[enum.Enum]):
+        """Gets the highest BadgeStatus.
+
+        This method works because `cls.__members__` is an ordered dict.
+
+        Args:
+            statuses: A list of BadgeStatuses.
```
```suggestion
            statuses: An interable of BadgeStatuses. The highest priority package in this iterable will be returned.
```

badge_server/main.py
```diff
@@ -76,6 +77,23 @@ class BadgeStatus(enum.Enum):
...
+
+        Returns:
+            A BadgeStatus. An error is thrown if no BadgeStatus exists.
```
brianquinlan: http://google.github.io/styleguide/pyguide.html#244-decision

```suggestion
            The BadgeStatus found in `statuses` that has the highest priority. For example, INTERNAL_ERROR would be returned if `statuses` contained INTERNAL_ERROR and OUTDATED_DEPENDENCY.
            
  Raises:
     ValueError: if `statuses` does not contain a valid BadgeStatus.
```

badge_server/main.py

badge_server/main.py

badge_server/main.py
```diff
@@ -76,6 +77,29 @@ class BadgeStatus(enum.Enum):
...
+        Args:
+            statuses: A list of BadgeStatuses.
+
+        Returns:
+            The BadgeStatus found in `statuses` that has the highest priority.
+            For example, INTERNAL_ERROR would be returned if `statuses`
+            contained INTERNAL_ERROR and OUTDATED_DEPENDENCY.
+
+        Raises:
+            ValueError if no BadgeStatus exists. For example, if the length of
```
```suggestion
            ValueError: if no BadgeStatus exists. For example, if the length of
```

PR #278, opened 2019-03-27
add logic and test for detecting missing packages

ylil93:


badge_server/main.py
```diff
@@ -104,6 +104,26 @@ class BadgeStatus(enum.Enum):
 }
 
 
+def _get_supported_versions(package_name: str) -> Iterable[int]:
```
```suggestion
def _get_supported_versions(package_name: str) -> List[int]:
```

`Iterable` there is not correct because you take the len() on line 146

badge_server/main.py
```diff
@@ -104,6 +104,26 @@ class BadgeStatus(enum.Enum):
 }
 
 
+def _get_supported_versions(package_name: str) -> List[int]:
+    """Gets the given package's supported python versions
+
+    Args:
+        package_name: The package name.
```
brianquinlan: Try to avoid studdering if your doc strings.
```suggestion
        package_name: The name of the package to return to the versions for e.g. "tensorflow"
```

PR #279, opened 2019-03-27
clean up _get_pair_compatibility_dict()

ylil93:
improve details when a package does not support a version of python

### DIFF
--- a/badge_server/test_badge_server.py
+++ b/badge_server/test_badge_server.py
@@ -73,8 +73,10 @@ class TestBadgeServer(unittest.TestCase):
 
     def test__get_pair_compatibility_dict_success(self):
         expected = {
-            'py2': {'status': main.BadgeStatus.SUCCESS, 'details': None},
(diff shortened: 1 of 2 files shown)


badge_server/test_badge_server.py
```diff
@@ -134,8 +136,9 @@ def test__get_pair_compatibility_dict_install_error(self):
         from compatibility_lib import package
 
         expected = {
-            'py2': { 'status': main.BadgeStatus.SUCCESS, 'details': {} },
-            'py3': { 'status': main.BadgeStatus.SUCCESS, 'details': {} },
+            'py2': {'status': main.BadgeStatus.SUCCESS,
+                    'details': 'The package does not support this version of python.'},
```
brianquinlan: That's kind of a weird success message.

brianquinlan: Maybe we need a different status for this? Where is this message used?

PR #282, opened 2019-03-29
Fix logic for finding missing data

ylil93:
- also fixes target page readability

### DIFF
--- a/badge_server/main.py
+++ b/badge_server/main.py
@@ -147,20 +147,24 @@ def _get_missing_details(package_names: List[str],
         package_names: A list of length 1 or 2 of the package name(s) to look
             up, e.g. ["tensorflow"], ["tensorflow", "opencensus"].
         results: A list of length 1 or 2 of the `CompatibilityResults` for the
-            given package(s). Results should come from datastore.
+            given package(s). The package names in the `CompatibilityResults`
+            should match the package_names argument.
 
     Returns:
-        None if there is no missing data; a description of the version(s)
-        missing otherwise, e.g. "Missing data for python version(s): 2".
+        None if there is no missing data; a description of the package(s) and
+        version(s) missing otherwise.
+        For example, if package_names=['opencensus'], and given that
+        `opencensus` is compatible with both python versions 2 and 3, if
+        `results` only contained a result for 3, this would be returned:
+        "Missing data for packages=['opencensus'], versions=[2]".
     """
-    bad_list_length = 'package_names: Expected length of 1 or 2, got {}'
-    not_whitelisted = 'One of the packages in {} is not whitelisted'
-
-    assert len(package_names) in (1, 2), bad_list_length.format(
+    message = 'package_names: Expected length of 1 or 2, got {}'.format(
         len(package_names))
+    assert len(package_names) in (1, 2), message
 
-    assert compat_utils._is_package_in_whitelist(
-        package_names), not_whitelisted.format(str(package_names))
+    message = 'One of the packages in {} is not whitelisted'.format(
+        str(package_names))
+    assert compat_utils._is_package_in_whitelist(package_names), message
 
     all_versions = (2, 3)
     versions_supported = set(all_versions)
--- a/badge_server/main.py
+++ b/badge_server/main.py
@@ -48,7 +48,7 @@ from compatibility_lib.dependency_highlighter import PriorityLevel
 
 from compatibility_lib import configs
 from compatibility_lib import package
-from typing import FrozenSet, Iterable, List
+from typing import FrozenSet, Iterable, List, Optional
 
 app = flask.Flask(__name__)
 
@@ -139,52 +139,53 @@ DEPENDENCY_STATUS_TO_BADGE_STATUS = {
 
 
 def _get_missing_details(package_names: List[str],
-                         results: Iterable[CompatibilityResult]) -> str:
+                         results: Iterable[CompatibilityResult]
+                         ) -> Optional[str]:
     """Gets the details for any missing data (if there is any)
 
     Args:
         package_names: A list of length 1 or 2 of the package name(s) to look
             up, e.g. ["tensorflow"], ["tensorflow", "opencensus"].
         results: A list of length 1 or 2 of the `CompatibilityResults` for the
-            given package(s).
+            given package(s). Results should come from datastore.
 
     Returns:
         None if there is no missing data; a description of the version(s)
         missing otherwise, e.g. "Missing data for python version(s): 2".
     """
     expected_lengths = (1, 2)
-    bad_list_length = 'package_names: Expected length of 1 or 2, got %s'
-    not_whitelisted = 'One of the packages in %s is not whitelisted'
+    bad_list_length = 'package_names: Expected length of 1 or 2, got {}'
+    not_whitelisted = 'One of the packages in {} is not whitelisted'
 
-    if len(package_names) not in expected_lengths:
-        # This should not happen
-        raise ValueError(bad_list_length % len(package_names))
+    assert len(package_names) in expected_lengths, bad_list_length.format(
+        len(package_names))
 
-    if not compat_utils._is_package_in_whitelist(package_names):
-        # This should not happen
-        raise ValueError(not_whitelisted % str(package_names))
+    assert compat_utils._is_package_in_whitelist(
+        package_names), not_whitelisted.format(str(package_names))
 
-    versions = (2, 3)
-    unsupported_package_mapping = configs.PKG_PY_VERSION_NOT_SUPPORTED
-    expected = {version: all([name not in unsupported_package_mapping[version]
-                              for name in package_names])
-                for version in versions}
+    all_versions = (2, 3)
 
-    actual = {version: False for version in versions}
-    for result in results:
-        actual[result.python_major_version] = True
+    versions_supported = set(version for version in all_versions)
+    for version in all_versions:
+        for package_name in package_names:
+            if package_name in configs.PKG_PY_VERSION_NOT_SUPPORTED[version]:
+                versions_supported.discard(version)
 
-    missing_versions = []
-    for version in versions:
-        if expected[version] and not actual[version]:
-            missing_versions.append(str(version))
+    versions_seen = {result.python_major_version for result in results}
 
-    if len(missing_versions) == 0:
+    if versions_seen.issuperset(versions_supported):
         return None
 
-    missing_details = 'Missing data for python version(s): {}'.format(
-        ' and '.join(missing_versions))
+    missing_versions = [v for v in versions_supported - versions_seen]
+    missing_details = 'Missing data for packages={}, versions={}'.format(
+        package_names, missing_versions)
     return missing_details
+
+    missing_details = 'Missing data for python version(s): {}'.format(
+        ' and '.join([str(v) for v in versions_supported - versions_seen]))
+    return missing_details
+
+
 
 
 def _get_self_compatibility_dict(package_name: str) -> dict:
--- a/badge_server/main.py
+++ b/badge_server/main.py
@@ -41,6 +41,7 @@ import pybadges
 
 import utils as badge_utils
 from compatibility_lib import utils as compat_utils
+from compatibility_lib.compatibility_store import CompatibilityResult
 from compatibility_lib.compatibility_store import Status as PackageStatus
 from compatibility_lib.dependency_highlighter import PriorityLevel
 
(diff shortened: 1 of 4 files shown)


badge_server/main.py
```diff
@@ -137,24 +138,53 @@ def get_highest_status(cls, statuses: Iterable[enum.Enum]):
...
+        raise ValueError(not_whitelisted % str(package_names))
+
+    versions = (2, 3)
     unsupported_package_mapping = configs.PKG_PY_VERSION_NOT_SUPPORTED
-    for version in (2, 3):
-        unsupported_packages = unsupported_package_mapping[version]
-        if package_name not in unsupported_packages:
-            supported.append(version)
-    return supported
+    expected = {version: all([name not in unsupported_package_mapping[version]
```
brianquinlan: Maybe this is more readable?

```python
package_supported_versions = {2, 3}
for version in [2, 3]:
  for package_name in package_names:
    if package_name in configs.PKG_PY_VERSION_NOT_SUPPORTED[version]:
      package_supported_versions.discard(version)

data_for_versions = {result.python_major_version for result in results}

if data_for_versions >= package_supported_versions:
  return None
else:
  return 'Missing compatibility for packages={}, versions={}'.format(package_names, package_supported_versions - data_for_versions)
```

badge_server/main.py
```diff
@@ -137,24 +138,53 @@ def get_highest_status(cls, statuses: Iterable[enum.Enum]):
...
+def _get_missing_details(package_names: List[str],
+                         results: Iterable[CompatibilityResult]) -> str:
+    """Gets the details for any missing data (if there is any)
 
     Args:
-        package_name: The name of the package to look up, e.g. "tensorflow".
+        package_names: A list of length 1 or 2 of the package name(s) to look
+            up, e.g. ["tensorflow"], ["tensorflow", "opencensus"].
+        results: A list of length 1 or 2 of the `CompatibilityResults` for the
+            given package(s).
```
brianquinlan: Could you say where the CompatibilityResults come from?

badge_server/templates/one-badge.html
```diff
@@ -173,14 +173,14 @@ <h4>Google Compatibility Check Results</h4>
                         <ul class="nav nav-tabs" id="myTab" role="tablist">
                             <li class="nav-item">
                                 <a class="nav-link active" id="py2-tab" data-toggle="tab" href="#py2" role="tab" aria-controls="py2" aria-selected="true">Python 2
-                                {% if google_compat_res.py2.status == "CHECK_WARNING" %}
+                                {% if "INCOMPATIBLE" in google_compat_res.py2.status.name %}
```
brianquinlan: Why "in"?

badge_server/main.py
```diff
@@ -137,24 +138,53 @@ def get_highest_status(cls, statuses: Iterable[enum.Enum]):
 }
 
 
-def _get_supported_versions(package_name: str) -> List[int]:
-    """Gets the given package's supported python versions
+def _get_missing_details(package_names: List[str],
+                         results: Iterable[CompatibilityResult]) -> str:
```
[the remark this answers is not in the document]

badge_server/main.py
```diff
@@ -137,24 +138,53 @@ def get_highest_status(cls, statuses: Iterable[enum.Enum]):
...
-    supported = []
-    if not compat_utils._is_package_in_whitelist([package_name]):
-        return supported
+    expected_lengths = (1, 2)
+    bad_list_length = 'package_names: Expected length of 1 or 2, got %s'
+    not_whitelisted = 'One of the packages in %s is not whitelisted'
+
+    if len(package_names) not in expected_lengths:
+        # This should not happen
+        raise ValueError(bad_list_length % len(package_names))
```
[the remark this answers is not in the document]

badge_server/main.py
```diff
@@ -137,24 +138,53 @@ def get_highest_status(cls, statuses: Iterable[enum.Enum]):
...
+
+    if not compat_utils._is_package_in_whitelist(package_names):
```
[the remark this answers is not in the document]

badge_server/main.py
```diff
@@ -137,24 +138,53 @@ def get_highest_status(cls, statuses: Iterable[enum.Enum]):
...
+        # This should not happen
+        raise ValueError(not_whitelisted % str(package_names))
+
+    versions = (2, 3)
```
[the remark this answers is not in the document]
ylil93: I chose to use a tuple, since the data inside would be static. Why use a list?

Note: I changed the variable name to `all_versions`

badge_server/main.py
ylil93: Done, but I made some minor tweaks to the variable names and used `issuperset()` instead of `>=` since that seemed clearer to me.

badge_server/main.py
ylil93: Done

badge_server/main.py
```diff
@@ -137,24 +138,54 @@ def get_highest_status(cls, statuses: Iterable[enum.Enum]):
...
-    supported = []
-    if not compat_utils._is_package_in_whitelist([package_name]):
-        return supported
-    unsupported_package_mapping = configs.PKG_PY_VERSION_NOT_SUPPORTED
-    for version in (2, 3):
-        unsupported_packages = unsupported_package_mapping[version]
-        if package_name not in unsupported_packages:
-            supported.append(version)
-    return supported
+    expected_lengths = (1, 2)
```
brianquinlan: I think that function-level constants are pretty rare in Python. You could ask Rebecca (readability reviewer) if you want a second opinion.

badge_server/main.py
```diff
@@ -137,24 +138,54 @@ def get_highest_status(cls, statuses: Iterable[enum.Enum]):
...
+
+    assert len(package_names) in expected_lengths, bad_list_length.format(
+        len(package_names))
+
+    assert compat_utils._is_package_in_whitelist(
+        package_names), not_whitelisted.format(str(package_names))
+
+    all_versions = (2, 3)
+
+    versions_supported = set(version for version in all_versions)
```
```suggestion
    versions_supported = set(all_versions)
```

badge_server/main.py
```diff
@@ -137,24 +138,54 @@ def get_highest_status(cls, statuses: Iterable[enum.Enum]):
...
+        for package_name in package_names:
+            if package_name in configs.PKG_PY_VERSION_NOT_SUPPORTED[version]:
+                versions_supported.discard(version)
+
+    versions_seen = {result.python_major_version for result in results}
+
+    if versions_seen.issuperset(versions_supported):
+        return None
+
+    missing_versions = [v for v in versions_supported - versions_seen]
```
brianquinlan: list(versions_support - versions_seen)
```suggestion
    missing_versions = list(versions_supported - versions_seen)
```

But I don't know why the conversion to list is necessary.

badge_server/templates/one-badge.html
ylil93: `CHECK WARNING` is now either `SELF INCOMPATIBLE` or `INCOMPATIBLE`

badge_server/main.py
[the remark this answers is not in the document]
brianquinlan: In Python, tuples tend to be used for heterogenous collections and list for homogeneous collections:
https://mail.python.org/pipermail/python-dev/2014-April/134140.html
https://github.com/google/styleguide/blob/gh-pages/pyguide.md#3199-tuples-vs-lists

badge_server/templates/one-badge.html
brianquinlan: I don't think that you should rely on that property.... on the other hand it is convenient. It's up to you.

badge_server/main.py
ylil93: The error message without it bothered me - I wanted the presentation to be consistent:
"Missing data for packages=['opencensus'], versions={2, 3}"

badge_server/main.py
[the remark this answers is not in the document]
ylil93: I honestly don't mind using a list instead, but this my first time hearing about homogeneity/heterogeneity being a distinction and I'd like determine if this should actually be a best practice or not because that would affect me going forward. Personally, the only distinction I've heard of is that tuples are immutable while lists are mutable. This discussion seems to back that up: https://bugs.python.org/issue14840 unless I'm misinterpreting things/am missing some other context.

Specifically these ones:
"Im/mutability is part of the definition and operation of the language. Homo/heter/geneity is not (that I can think of at the moment)."

"there are parts of Python's own syntax that
require a tuple that happens to be a homogenous sequence, such as the
:keyword:`except` clause of a ``try ... except`` statement, :func:`issubclass`,
and :func:`isinstance`"

badge_server/main.py

badge_server/main.py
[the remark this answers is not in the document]
brianquinlan: OK, I think that I retract my comment - both patterns seem equally common in Google3. I started a discussion in the team chat ;-)

badge_server/main.py
```diff
@@ -137,24 +138,46 @@ def get_highest_status(cls, statuses: Iterable[enum.Enum]):
...
+                         results: Iterable[CompatibilityResult]
+                         ) -> Optional[str]:
+    """Gets the details for any missing data (if there is any)
 
     Args:
-        package_name: The name of the package to look up, e.g. "tensorflow".
+        package_names: A list of length 1 or 2 of the package name(s) to look
+            up, e.g. ["tensorflow"], ["tensorflow", "opencensus"].
+        results: A list of length 1 or 2 of the `CompatibilityResults` for the
+            given package(s). Results should come from datastore.
```
brianquinlan: I mean specifically that they should be single or pairwise compatibility results i.e. they should be for the same set of packages. You could instead write something like:

"The package names in the CompatibilityResults should match the `package_names` argument"

badge_server/main.py
```diff
@@ -137,24 +138,46 @@ def get_highest_status(cls, statuses: Iterable[enum.Enum]):
...
 
     Returns:
-        A list of supported python versions.
+        None if there is no missing data; a description of the version(s)
+        missing otherwise, e.g. "Missing data for python version(s): 2".
```
brianquinlan: Use an actual string example.

badge_server/main.py
```diff
@@ -137,24 +138,46 @@ def get_highest_status(cls, statuses: Iterable[enum.Enum]):
...
-    supported = []
-    if not compat_utils._is_package_in_whitelist([package_name]):
-        return supported
-    unsupported_package_mapping = configs.PKG_PY_VERSION_NOT_SUPPORTED
-    for version in (2, 3):
-        unsupported_packages = unsupported_package_mapping[version]
-        if package_name not in unsupported_packages:
-            supported.append(version)
-    return supported
+    bad_list_length = 'package_names: Expected length of 1 or 2, got {}'
```
brianquinlan: I still don't like these function-level constants ;-)

badge_server/main.py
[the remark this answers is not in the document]
ylil93: I don't think I'm in the team chat.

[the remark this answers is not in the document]
brianquinlan: I just added you.

badge_server/templates/one-badge.html
ylil93: To give more context, this code inserts a 'warning' message if that statement is true, which looks like this:
warning

And I think refactoring the UI is beyond the scope of this PR (the goal is just to fix things that got broken).

badge_server/main.py
```diff
@@ -42,12 +42,13 @@
 
 import utils as badge_utils
 from compatibility_lib import utils as compat_utils
+from compatibility_lib.compatibility_store import CompatibilityResult
```
brianquinlan: You can only import modules (see http://google.github.io/styleguide/pyguide.html#22-imports)

badge_server/main.py
ylil93: Done

badge_server/main.py
ylil93: Oh ok got it - done

badge_server/main.py
ylil93: changed it

badge_server/main.py

badge_server/main.py
ylil93: removed

badge_server/main.py
```diff
@@ -42,12 +42,12 @@
 
 import utils as badge_utils
 from compatibility_lib import utils as compat_utils
-from compatibility_lib.compatibility_store import Status as PackageStatus
-from compatibility_lib.dependency_highlighter import PriorityLevel
+from compatibility_lib import dependency_highlighter as highlighter
```
brianquinlan: Unfortunately, you can't do this either:
https://github.com/google/styleguide/blob/gh-pages/pyguide.md#224-decision

ylil93: I think at >20 characters, 'dependency_highlighter' would count as inconveniently long?

PR #283, opened 2019-03-29
Use the PRD spelling for the badge RHS.

brianquinlan:


badge_server/main.py
```diff
@@ -112,6 +112,17 @@ def get_highest_status(cls, statuses: Iterable[enum.Enum]):
     BadgeStatus.SUCCESS: 'brightgreen',
 }
 
+BADGE_STATUS_TO_TEXT = {
```
ylil93: Can't these be the direct values of `BadgeStatus`?

brianquinlan: They could be. It seems like this is a more clean separation though.

ylil93: Hmm... it just seemed a bit redundant since the enum values for BadgeStatus are constant strings anyway. I think having the `_badge_status_to_text()` function would be sufficient for clarity?

brianquinlan: OK, fixed.

PR #285, opened 2019-04-02
Use timezone-unaware UTC times everywhere

brianquinlan:


ylil93:
Do any tests need to be updated?

PR #287, opened 2019-04-04
Add a more complete framework for badge image testing

brianquinlan:


badge_server/main.py
```diff
@@ -448,14 +448,20 @@ def one_badge_image():
...
+
+    badge_args = dict(
         left_text=badge_name,
         right_text=_badge_status_to_text(status),
         right_color=color,
         whole_link=details_link)
 
-    response = flask.make_response(badge)
-    response.content_type = badge_utils.SVG_CONTENT_TYPE
+    if flask.request.accept_mimetypes.best_match(
```
ylil93: I haven't used flask very much - can you explain (1) what this check does and (2) why this if/else statement is better than what was there before?

badge_server/test_badges.py
```diff
@@ -0,0 +1,206 @@
...
+    [
+        package.Package('google-api-python-client'),
+        package.Package('tensorflow')
+    ],
+    python_major_version=3,
+    status=compatibility_store.Status.SUCCESS,
+    timestamp=datetime.datetime.utcnow())
+
+
+class TestBadgeServer(unittest.TestCase):
```
ylil93: Seems like these tests should belong in test_badge_server.py?

brianquinlan: I think that these tests will end up replacing those and I was imagining splitting the tests into two modules: one for the image output and one for the HTML output.

badge_server/main.py
brianquinlan: I changed it - let me know if this is more clear.

ylil93: The change makes it way more clear! Thanks!

badge_server/test_badges.py
ylil93: ok sgtm

PR #289, opened 2019-04-06
Fix a bug where package version handling wasn't correct for github packages

brianquinlan:


badge_server/test_badges.py
```diff
@@ -120,13 +231,22 @@ def setUp(self):
...
+            'compatibility_lib.configs.PKG_LIST', [
+                'apache-beam[gcp]',
+                'google-api-core',
+                'google-api-python-client',
+                'tensorflow',
+            ])
+        self._whitelist_urls_patch = unittest.mock.patch(
+            'compatibility_lib.configs.WHITELIST_URLS', {
+                'git+git://github.com/google/apache-beam.git':
+                'apache-beam[gcp]',
```
ylil93: This needs to be indented

http://google.github.io/styleguide/pyguide.html#34-indentation

badge_server/test_badges.py
```diff
@@ -120,13 +231,22 @@ def setUp(self):
...
+                'git+git://github.com/google/api-core.git': 'google-api-core',
+                'git+git://github.com/google/api-python-client.git':
+                'google-api-python-client',
```
ylil93: This needs to be indented

http://google.github.io/styleguide/pyguide.html#34-indentation

badge_server/test_badges.py
```diff
@@ -135,74 +255,106 @@ def setUp(self):
...
-            GOOGLE_API_PYTHON_CLIENT_TENSORFLOW_RECENT_SUCCESS_3,
-        ])
-        json_response = self._get_image_json('google-api-core')
+            '/one_badge_image', query_string={
+                'package': package
+            }).get_json()
+
+    def assertLinkUrl(self, package, actual_url):
+        """Assert that the link for the badge image is correct for a package."""
+        o = urllib.parse.urlparse(actual_url)
```
ylil93: What does `o` mean?

badge_server/test_badges.py
brianquinlan: Thanks! I don't know why yapf did this - maybe I forgot to use --style=google?

badge_server/test_badges.py
brianquinlan: Done

badge_server/test_badges.py
brianquinlan: It was the example used in the docs...I fixed it.

PR #290, opened 2019-04-08
UI fixes

ylil93:
- use exact (hex colors) over named ones due to discrepancy between shields io and css
- generate css based on BADGE_STATUS_TO_COLOR
- update tests

### DIFF
--- a/badge_server/static/css/theme.css
+++ b/badge_server/static/css/theme.css
@@ -179,7 +179,6 @@
     margin-bottom: 0;
 }
 
-/* Clear browser cache: cmd + shift + R  */
 .status {
     display: inline-block;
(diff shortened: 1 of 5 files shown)


badge_server/static/css/theme.css
```diff
@@ -200,6 +179,7 @@
     margin-bottom: 0;
 }
 
+/* Clear browser cache: cmd + shift + R  */
```
brianquinlan: What's this comment about?

ylil93: Oh it's a shortcut for clearing browser cache, which can contain stale css that doesn't reflect recent changes made, but since this isn't actually related to any code, I can remove it.

PR #294, opened 2019-04-09
 expand test coverage

ylil93:
- the case where at least one package in a pair check does not support the given version

### DIFF
--- a/badge_server/test_badges.py
+++ b/badge_server/test_badges.py
@@ -390,17 +390,14 @@ class TestBadgeImageSuccess(BadgeImageTestCase):
         self.assertEqual(json_response['right_color'], '#44CC44')
         self.assertLinkUrl(package, json_response['whole_link'])
 
-    def test_pypi_mix_fresh_nodeps(self):
-        RECENT_SUCCESS_DATA_EXTRA = RECENT_SUCCESS_DATA + [
+    def test_pypi_py2py3_fresh_nodeps(self):
+        fake_results = RECENT_SUCCESS_DATA + [
             APACHE_BEAM_GOOGLE_API_CORE_RECENT_INSTALL_ERROR_3,
             APACHE_BEAM_GOOGLE_API_CORE_GIT_RECENT_INSTALL_ERROR_3,
-            APACHE_BEAM_GIT_GOOGLE_API_CORE_RECENT_INSTALL_ERROR_3,
             GOOGLE_API_CORE_TENSORFLOW_RECENT_INSTALL_ERROR_2,
-            GOOGLE_API_CORE_TENSORFLOW_GIT_RECENT_INSTALL_ERROR_2,
-            GOOGLE_API_CORE_TENSORFLOW_GIT_RECENT_SUCCESS_3,
             GOOGLE_API_CORE_GIT_TENSORFLOW_RECENT_INSTALL_ERROR_2,
         ]
-        self.fake_store.save_compatibility_statuses(RECENT_SUCCESS_DATA_EXTRA)
+        self.fake_store.save_compatibility_statuses(fake_results)
         package = 'google-api-core'
         json_response = self.get_image_json(package)
         self.assertEqual(json_response['left_text'],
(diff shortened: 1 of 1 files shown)


badge_server/test_badges.py
```diff
@@ -334,6 +422,52 @@ def test_git_py3_fresh_nodeps(self):
         self.assertEqual(json_response['right_color'], '#44CC44')
         self.assertLinkUrl(package, json_response['whole_link'])
 
+    def test_pypi_mix_fresh_nodeps(self):
+        RECENT_SUCCESS_DATA_EXTRA = RECENT_SUCCESS_DATA + [
```
brianquinlan: I'm not sure if this data setup is interesting because you'd get the same result even if that continue statement wasn't there.

You probably need something like this (just an outline):
```python
APACHE_BEAM_GOOGLE_API_CORE_RECENT_FAILURE_3 = compatibility_store.CompatibilityResult(
    [
        package.Package('apache-beam[gcp]'),
        package.Package('google-api-core')
    ],
    python_major_version=3,  # apache-beam does not support Python 3
    status=compatibility_store.Status.FAILURE,
    timestamp=datetime.datetime.utcnow())

    def test_pypi_py2py3_fresh_no_deps_failure_in_unsupported_version():
       self.fake_store.save_compatibility_statuses(
            RECENT_SUCCESS_DATA + 
            [APACHE_BEAM_GOOGLE_API_CORE_RECENT_FAILURE_3])
       # rest of your test is fine, also need to do the same for github.

```

ylil93: It's expected to get a `SUCCESS` status if the continue statement gets hit; it's there to make sure we don't compare CompatibilityResults for versions that aren't supported, so I'm not sure if there is a way to make this test more interesting.

brianquinlan: I mean that this test will pass even without the continue statement, right? (you could try it). A test for a particular behavior  isn't very convincing if the code could not even be there and it still passes ;-)

My variation will fail without the continue statement.

ylil93: OH ok I misread your example. I get what you meant now.

ylil93: Done

badge_server/test_badges.py
```diff
@@ -82,6 +82,13 @@
     status=compatibility_store.Status.SUCCESS,
     timestamp=datetime.datetime.utcnow())
 
+APACHE_BEAM_GOOGLE_API_CORE_RECENT_SUCCESS_3 = compatibility_store.CompatibilityResult(
```
brianquinlan: The name has "SUCCESS" in it.

badge_server/test_badges.py
```diff
@@ -98,6 +105,26 @@
     status=compatibility_store.Status.SUCCESS,
     timestamp=datetime.datetime.utcnow())
 
+APACHE_BEAM_GIT_GOOGLE_API_CORE_RECENT_SUCCESS_3 = compatibility_store.CompatibilityResult(
```
brianquinlan: "SUCCESS"

badge_server/test_badges.py
brianquinlan: Do you need all of the constants below? For the tests that you added, don't need you just need two constants?

ylil93: I thought it would be good to get all of the permutations, but if you don't think that's necessary, I can remove some.

badge_server/test_badges.py
ylil93: renamed all the constants

badge_server/test_badges.py
```diff
@@ -98,6 +105,26 @@
...
 
+APACHE_BEAM_GIT_GOOGLE_API_CORE_RECENT_INSTALL_ERROR_3 = compatibility_store.CompatibilityResult(
+    [package.Package('git+git://github.com/google/apache-beam.git'),
+     package.Package('google-api-core')],
+    python_major_version=3,   # apache-beam does not support Python 3
+    status=compatibility_store.Status.INSTALL_ERROR,
+    timestamp=datetime.datetime.utcnow())
+
+APACHE_BEAM_GIT_GOOGLE_API_CORE_GIT_RECENT_SUCCESS_2 = compatibility_store.CompatibilityResult(
+    [package.Package('git+git://github.com/google/apache-beam.git'),
```
brianquinlan: We never do this check, right? I thought that we never check git against git?

ylil93: removed

badge_server/test_badges.py
```diff
@@ -334,6 +390,76 @@ def test_git_py3_fresh_nodeps(self):
         self.assertEqual(json_response['right_color'], '#44CC44')
         self.assertLinkUrl(package, json_response['whole_link'])
 
+    def test_pypi_py2py3_fresh_nodeps(self):
```
brianquinlan: You need to pick names to indicate what you are testing. Maybe:
`test_pypi_py2py3_fresh_nodeps_failure_in_unsupported_version_combination`
   "Tests that failures in pair compatibility results for unsupported Python versions are ignored." 

But too long...

   """Tests that irrelevant pair compatibility results are ignored.

   Pair compatibility results for unsupported Python versions should be ignored.
   For example, the pair <apache-beam, google-api-core> for Python 3 should
   be ignored because apache-beam does not support Python 3. 
   """

brianquinlan: Actually, what is this testing? And this duplicates an existing name?

ylil93: This used to be `test_pypi_mix_fresh_nodeps` but you mentioned in an earlier comment that `mix` should be changed to `py2py3` but I didn't realize that would cause a naming conflict. I changed the name to something self evident.

PR #295, opened 2019-04-10
Use the correct type for failure details

ylil93:
- make sure status is a dict for pairwise result when handling an internal error


compatibility_lib/compatibility_lib/compatibility_store.py
```diff
@@ -232,6 +232,9 @@ def __init__(self,
         # are None.
         if mysql_host is None and mysql_unix_socket is None:
             mysql_host = '127.0.0.1'
+            assert mysql_user is not None
```
brianquinlan: Could you put this unrelated change in another PR?

ylil93: done

PR #296, opened 2019-04-11
Remove the ability to mutate the timestamp of a CompatibilityResult

brianquinlan:


ylil93:
Next time can you make sure you're even with the upstream/master branch before making new changes? The change is small in this case so it's ok, but having old commit histories in the PR makes it confusing to review.

PR #305, opened 2019-04-13
Add a new method that allows dependency information in CompatibilityResults to be updated.

brianquinlan:


compatibility_lib/compatibility_lib/compatibility_store.py
```diff
@@ -97,6 +97,44 @@ def __eq__(self, o):
                     self.timestamp == o.timestamp)
         return NotImplemented
 
+    def with_updated_dependency_info(
```
ylil93: Would this be used in production code or would this be beneficial for testing only?

brianquinlan: It could be used for production code but right now I want to use it for testing.

ylil93: If that's the case, I think it makes more sense to house this method in our fake store or as a utility function in the test suite.

brianquinlan: I think that this method is logical for this class i.e. it makes sense for both test and non-test usage. Right now I plan on using it for tests but it would be equally useful outside of tests.

ylil93: But you just said that you have no plans to use it outside of testing, right? I just find it strange to add this code to production if that's the case.

brianquinlan: I think that your argument makes sense but I'm not fully convinced ;-)

I think that this method belongs with the class just like `__repr__` (which is usually only used for debug) or any other method.

ylil93: I'm punt this to the rest of the team, since I'm not entirely convinced by that argument either.

ylil93: Given the response on the thread, I'm comfortable with approving these changes!

brianquinlan: I can change the approach if there is more feedback.

PR #307, opened 2019-04-22
add dependency tests

ylil93:
- add tests for old dependencies
- add tests for obsolete dependencies

### DIFF
--- a/badge_server/test_badges.py
+++ b/badge_server/test_badges.py
@@ -631,9 +631,16 @@ class TestBadgeImageOutdatedDependency(BadgeImageTestCase):
 
     def test_pypi_py2py3_off_by_minor(self):
         old_dep_info = dict(UP_TO_DATE_DEPS)
-        old_dep_info['google-auth'] = dict(UP_TO_DATE_DEPS['google-auth'])
-        old_dep_info['google-auth']['installed_version'] = '1.4.0'
-        old_dep_info['google-auth']['is_latest'] = False
+        old_dep_info['google-auth'] = {
+            'current_time': datetime.datetime.utcnow(),
+            'installed_version': '1.4.0',
+            'installed_version_time': datetime.datetime(
+                2019, 2, 19, 21, 15, 56),
+            'is_latest': False,
+            'latest_version': '1.6.3',
+            'latest_version_time': datetime.datetime(
+                2019, 2, 19, 21, 15, 56)
+        }
 
         old_dep_compat_results = list(RECENT_SUCCESS_DATA)
         old_dep_compat_results.remove(GOOGLE_API_CORE_RECENT_SUCCESS_2)
@@ -658,9 +665,16 @@ class TestBadgeImageOutdatedDependency(BadgeImageTestCase):
 
     def test_git_py2py3_off_by_minor(self):
         old_dep_info = dict(UP_TO_DATE_DEPS)
-        old_dep_info['google-auth'] = dict(UP_TO_DATE_DEPS['google-auth'])
-        old_dep_info['google-auth']['installed_version'] = '1.4.0'
-        old_dep_info['google-auth']['is_latest'] = False
+        old_dep_info['google-auth'] = {
+            'current_time': datetime.datetime.utcnow(),
+            'installed_version': '1.4.0',
+            'installed_version_time': datetime.datetime(
+                2019, 2, 19, 21, 15, 56),
+            'is_latest': False,
+            'latest_version': '1.6.3',
+            'latest_version_time': datetime.datetime(
+                2019, 2, 19, 21, 15, 56)
+        }
 
         old_dep_compat_results = list(RECENT_SUCCESS_DATA)
         old_dep_compat_results.remove(GOOGLE_API_CORE_GIT_RECENT_SUCCESS_2)
@@ -685,9 +699,16 @@ class TestBadgeImageOutdatedDependency(BadgeImageTestCase):
 
     def test_pypi_py2py3_off_by_patch(self):
         old_dep_info = dict(UP_TO_DATE_DEPS)
-        old_dep_info['google-auth'] = dict(UP_TO_DATE_DEPS['google-auth'])
-        old_dep_info['google-auth']['installed_version'] = '1.6.0'
-        old_dep_info['google-auth']['is_latest'] = False
+        old_dep_info['google-auth'] = {
+            'current_time': datetime.datetime.utcnow(),
+            'installed_version': '1.6.0',
+            'installed_version_time': datetime.datetime(
+                2019, 2, 19, 21, 15, 56),
+            'is_latest': False,
+            'latest_version': '1.6.3',
+            'latest_version_time': datetime.datetime(
+                2019, 2, 19, 21, 15, 56)
+        }
 
         old_dep_compat_results = list(RECENT_SUCCESS_DATA)
         old_dep_compat_results.remove(GOOGLE_API_CORE_RECENT_SUCCESS_2)
@@ -712,9 +733,16 @@ class TestBadgeImageOutdatedDependency(BadgeImageTestCase):
 
     def test_git_py2py3_off_by_patch(self):
         old_dep_info = dict(UP_TO_DATE_DEPS)
-        old_dep_info['google-auth'] = dict(UP_TO_DATE_DEPS['google-auth'])
-        old_dep_info['google-auth']['installed_version'] = '1.6.0'
-        old_dep_info['google-auth']['is_latest'] = False
+        old_dep_info['google-auth'] = {
+            'current_time': datetime.datetime.utcnow(),
+            'installed_version': '1.6.0',
+            'installed_version_time': datetime.datetime(
+                2019, 2, 19, 21, 15, 56),
+            'is_latest': False,
+            'latest_version': '1.6.3',
+            'latest_version_time': datetime.datetime(
+                2019, 2, 19, 21, 15, 56)
+        }
 
         old_dep_compat_results = list(RECENT_SUCCESS_DATA)
         old_dep_compat_results.remove(GOOGLE_API_CORE_GIT_RECENT_SUCCESS_2)
@@ -742,9 +770,16 @@ class TestBadgeImageObsoleteDependency(BadgeImageTestCase):
 
     def test_pypi_py2py3_off_by_major(self):
         obsolete_dep_info = dict(UP_TO_DATE_DEPS)
-        obsolete_dep_info['google-auth'] = dict(UP_TO_DATE_DEPS['google-auth'])
-        obsolete_dep_info['google-auth']['installed_version'] = '0.9.9'
-        obsolete_dep_info['google-auth']['is_latest'] = False
+        obsolete_dep_info['google-auth'] = {
+            'current_time': datetime.datetime.utcnow(),
+            'installed_version': '0.9.9',
+            'installed_version_time': datetime.datetime(
+                2019, 2, 19, 21, 15, 56),
+            'is_latest': False,
+            'latest_version': '1.6.3',
+            'latest_version_time': datetime.datetime(
+                2019, 2, 19, 21, 15, 56)
+        }
 
         obsolete_dep_compat_results = list(RECENT_SUCCESS_DATA)
         obsolete_dep_compat_results.remove(GOOGLE_API_CORE_RECENT_SUCCESS_2)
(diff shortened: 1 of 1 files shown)


badge_server/test_badges.py
```diff
@@ -593,5 +626,358 @@ def test_github_py2py3_py3_incompatible_fresh_nodeps(self):
         self.assertEqual(json_response['left_text'],
                          'compatibility check (master)')
         self.assertEqual(json_response['right_text'], 'self incompatible')
+
+class TestBadgeImageOutdatedDependency(BadgeImageTestCase):
+
+    def test_pypi_py2py3_off_by_minor(self):
+        old_dep_info = dict(UP_TO_DATE_DEPS)
+        old_dep_info['google-auth'] = dict(UP_TO_DATE_DEPS['google-auth'])
```
brianquinlan: This is just a thought...maybe it makes sense to just add this single old dependency to the compatibility result rather than partially re-writing the existing dependency. It just saves have to look at the constant to see what changed. e.g.

```
old_dep = {
        'current_time': datetime.datetime.utcnow(),
        'installed_version': '1.4.0',
        'installed_version_time': datetime.datetime(
            2019, 2, 19, 21, 15, 56),
        'is_latest': False,
        'latest_version': '1.4.0',
        'latest_version_time': datetime.datetime(
            2019, 2, 19, 21, 15, 56)
}
old_dep_compat_results = list(RECENT_SUCCESS_DATA)
old_dep_compat_results.remove(GOOGLE_API_CORE_RECENT_SUCCESS_2)
old_dep_compat_results.remove(GOOGLE_API_CORE_RECENT_SUCCESS_3)
old_dep_compat_results.append(
            GOOGLE_API_CORE_RECENT_SUCCESS_2.with_updated_dependency_info(
                old_dep_info))
old_dep_compat_results.append(
            GOOGLE_API_CORE_RECENT_SUCCESS_3.with_updated_dependency_info(
                old_dep_info))
```

I think that the existing compatibility results in `RECENT_SUCCESS_DATA` should also include dependency info, at least for the package itself (because that is the actual behavior).

badge_server/test_badges.py
```diff
@@ -593,5 +626,358 @@ def test_github_py2py3_py3_incompatible_fresh_nodeps(self):
         self.assertEqual(json_response['left_text'],
                          'compatibility check (master)')
         self.assertEqual(json_response['right_text'], 'self incompatible')
+
+class TestBadgeImageOutdatedDependency(BadgeImageTestCase):
```
brianquinlan: It can be in another PR, but we also need to test the success case with dependencies.

badge_server/test_badges.py
```diff
@@ -266,14 +266,47 @@
...
+        'installed_version': '1.9.0',
+        'installed_version_time': datetime.datetime(
+            2019, 4, 5, 18, 1, 48),
+        'is_latest': True,
+        'latest_version': '1.9.0',
+        'latest_version_time': datetime.datetime(
+            2019, 4, 5, 18, 1, 48)
+    },
+}
+
```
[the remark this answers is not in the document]
ylil93: There are - 301, 302

badge_server/test_badges.py
ylil93: Good point - will do in a follow up PR

ylil93: Actually I think that this case will be covered once dependencies are added to the CRs in `RECENT_SUCCESS_DATA`

badge_server/test_badges.py
ylil93: I agree that rewriting the entire dependency instead of partially would be more clear - I will start updating the test cases.

With regards to `RECENT_SUCCESS_DATA` including dep info, I'm fine with that being implemented, but probably not in this PR.

badge_server/test_badges.py
brianquinlan: I would imagine that the success case will have no have non-trivial dependencies - I would just add self dependencies to `RECENT_SUCCESS_DATA` since the data is not correct without it.

badge_server/test_badges.py
```diff
@@ -266,14 +266,47 @@
...
+        'current_time': datetime.datetime.utcnow(),
+        'installed_version': '1.19.0',
+        'installed_version_time': datetime.datetime(
+            2019, 2, 27, 0, 0, 53),
+        'is_latest': True,
+        'latest_version': '1.19.0',
+        'latest_version_time': datetime.datetime(
+            2019, 2, 27, 0, 0, 53)
+    },
+    'google-api-core': {
```
brianquinlan: Don't use a dependency that we also test.
```suggestion
    'requests': {
```

badge_server/test_badges.py
```diff
@@ -593,5 +626,407 @@ def test_github_py2py3_py3_incompatible_fresh_nodeps(self):
         self.assertEqual(json_response['left_text'],
                          'compatibility check (master)')
         self.assertEqual(json_response['right_text'], 'self incompatible')
+
+
+class TestBadgeImageOutdatedDependency(BadgeImageTestCase):
+    """Tests for cases where the badge image displays 'Outdated Dependency.'"""
```
brianquinlan:
```suggestion
    """Tests for cases where the badge image displays 'old dependency'."""
```

badge_server/test_badges.py
```diff
@@ -593,5 +626,407 @@ def test_github_py2py3_py3_incompatible_fresh_nodeps(self):
...
+        json_response = self.get_image_json(package_name)
+        self.assertEqual(json_response['left_text'],
+                         'compatibility check (master)')
+        self.assertEqual(json_response['right_text'], 'old dependency')
+        self.assertEqual(json_response['right_color'], '#A4A61D')
+        self.assertLinkUrl(package_name, json_response['whole_link'])
+
+
+class TestBadgeImageObsoleteDependency(BadgeImageTestCase):
+    """Tests for cases where the badge image displays 'Obsolete Dependency.'"""
```
brianquinlan:
```suggestion
    """Tests for cases where the badge image displays 'obsolete dependency'."""
```

badge_server/test_badges.py
```diff
@@ -593,5 +626,407 @@ def test_github_py2py3_py3_incompatible_fresh_nodeps(self):
...
+        package_name = 'git+git://github.com/google/api-core.git'
+        json_response = self.get_image_json(package_name)
+        self.assertEqual(json_response['left_text'],
+                         'compatibility check (master)')
+        self.assertEqual(json_response['right_text'], 'obsolete dependency')
+        self.assertEqual(json_response['right_color'], '#E05D44')
+        self.assertLinkUrl(package_name, json_response['whole_link'])
+
+    def test_pypi_py2py3_expired_major_grace_period(self):
+        obsolete_dep_info = dict(UP_TO_DATE_DEPS)
```
brianquinlan:
```suggestion
        """Tests that "old dependency" eventually changes to "obsolete ..."."""
        obsolete_dep_info = dict(UP_TO_DATE_DEPS)
```

badge_server/test_badges.py
```diff
@@ -593,5 +626,407 @@ def test_github_py2py3_py3_incompatible_fresh_nodeps(self):
...
+        package_name = 'google-api-core'
+        json_response = self.get_image_json(package_name)
+        self.assertEqual(json_response['left_text'],
+                         'compatibility check (PyPI)')
+        self.assertEqual(json_response['right_text'], 'obsolete dependency')
+        self.assertEqual(json_response['right_color'], '#E05D44')
+        self.assertLinkUrl(package_name, json_response['whole_link'])
+
+    def test_git_py2py3_expired_major_grace_period(self):
+        obsolete_dep_info = dict(UP_TO_DATE_DEPS)
```
brianquinlan:
```suggestion
        """Tests that "old dependency" eventually changes to "obsolete ..."."""
        obsolete_dep_info = dict(UP_TO_DATE_DEPS)
```

badge_server/test_badges.py
```diff
@@ -593,5 +626,407 @@ def test_github_py2py3_py3_incompatible_fresh_nodeps(self):
...
+        package_name = 'git+git://github.com/google/api-core.git'
+        json_response = self.get_image_json(package_name)
+        self.assertEqual(json_response['left_text'],
+                         'compatibility check (master)')
+        self.assertEqual(json_response['right_text'], 'obsolete dependency')
+        self.assertEqual(json_response['right_color'], '#E05D44')
+        self.assertLinkUrl(package_name, json_response['whole_link'])
+
+    def test_pypi_py2py3_expired_default_grace_period(self):
+        obsolete_dep_info = dict(UP_TO_DATE_DEPS)
```
brianquinlan:
```suggestion
        """Tests that "old dependency" eventually changes to "obsolete ..."."""
        obsolete_dep_info = dict(UP_TO_DATE_DEPS)
```

badge_server/test_badges.py
```diff
@@ -593,5 +626,407 @@ def test_github_py2py3_py3_incompatible_fresh_nodeps(self):
...
+        package_name = 'google-api-core'
+        json_response = self.get_image_json(package_name)
+        self.assertEqual(json_response['left_text'],
+                         'compatibility check (PyPI)')
+        self.assertEqual(json_response['right_text'], 'obsolete dependency')
+        self.assertEqual(json_response['right_color'], '#E05D44')
+        self.assertLinkUrl(package_name, json_response['whole_link'])
+
+    def test_git_py2py3_expired_default_grace_period(self):
+        obsolete_dep_info = dict(UP_TO_DATE_DEPS)
```
brianquinlan:
```suggestion
        """Tests that "old dependency" eventually changes to "obsolete ..."."""
        obsolete_dep_info = dict(UP_TO_DATE_DEPS)
```

badge_server/test_badges.py
ylil93: Without the 'google-api-core' dependency info, an error is triggered: https://github.com/GoogleCloudPlatform/cloud-opensource-python/blob/70b41cadaf7532cf03cee2e0bcc16f2706c0c5b7/compatibility_lib/compatibility_lib/compatibility_store.py#L194

(_This_ is case that I got mixed up with earlier today).

ylil93: `ValueError: missing version information for google-api-core`
Is the complete message

badge_server/test_badges.py
ylil93: Done

badge_server/test_badges.py
ylil93: Done

badge_server/test_badges.py
ylil93: Done

badge_server/test_badges.py
ylil93: Done

badge_server/test_badges.py
ylil93: Done

badge_server/test_badges.py
ylil93: Done

badge_server/test_badges.py
ylil93: In lieu of your discovery that a package must list itself as a dependency, this makes sense since the package being looked up is `google-api-core`.

brianquinlan: OK, I'd keep this dependency (and also add `requests`) but add a TODO saying that `google-api-core` should be removed when dependencies are properly added to the existing compatibility results

ylil93: My understanding was that `requests` was an arbitrary replacement for `google-api-core`. But given that I already have dummy data for other packages (`google-auth`, `grpcio`) why is it still necessary to add dummy data for `requests` in particular?

ylil93: Done

badge_server/test_badges.py
```diff
@@ -786,5 +822,411 @@ def test_github_py2py3_py3_incompatible_fresh_nodeps(self):
...
+        json_response = self.get_image_json(package_name)
+        self.assertEqual(json_response['left_text'],
+                         'compatibility check (master)')
+        self.assertEqual(json_response['right_text'], 'old dependency')
+        self.assertEqual(json_response['right_color'], '#A4A61D')
+        self.assertLinkUrl(package_name, json_response['whole_link'])
+
+
+class TestBadgeImageObsoleteDependency(BadgeTestCase):
+    """Tests for cases where the badge image displays 'obsolete dependency.'"""
```
brianquinlan:
```suggestion
    """Tests for cases where the badge image displays 'obsolete dependency'."""
```

Super nit ;-)

ylil93: Oh it's muscle memory for me to to put the punctuation inside the quotations.

PR #308, opened 2019-05-03
add dep info to existing CRs

ylil93:
add `dependency_info` to existing `CompatibilityResult`s in `test_badges`

### DIFF
--- a/badge_server/test_badges.py
+++ b/badge_server/test_badges.py
@@ -33,7 +33,7 @@ APACHE_BEAM_RECENT_SUCCESS_2 = compatibility_store.CompatibilityResult(
     status=compatibility_store.Status.SUCCESS,
     dependency_info={
         'apache-beam[gcp]': {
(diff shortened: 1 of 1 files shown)


badge_server/test_badges.py
```diff
@@ -31,72 +31,216 @@
...
+    },
     timestamp=datetime.datetime.utcnow())
 
 GOOGLE_API_CORE_RECENT_SUCCESS_2 = compatibility_store.CompatibilityResult(
     [package.Package('google-api-core')],
     python_major_version=2,
     status=compatibility_store.Status.SUCCESS,
+    dependency_info={
+        'google-api-core': {
+            'current_time': datetime.datetime.utcnow(),
```
adamcataldo: Is there a reason why this has to be "now"? This means that every time the test is run, it will be run with different input, which means it will be hard to reproduce. I see we've already been using this pattern, e.g. on old line 46.

PR #311, opened 2019-05-07
Change badge template to support new dependency status name

### LINKED ISSUE #310: Dependency check result isn't showing the details
Looks like the template file isn't being updated to use the new json format result.

liyanhui1228:
Fixes #310

### DIFF
--- a/badge_server/templates/one-badge.html
+++ b/badge_server/templates/one-badge.html
@@ -108,7 +108,7 @@
                                 <i class="zmdi zmdi-account-calendar"></i>Dependency Check Result: <span class="status {{ dependency_res.status.name | lower }}">{{ dependency_res.status.value }}</span>
                             </h3>
                             <div class="table-responsive table-data">
-                                {% if dependency_res.status.name in ['LOW_PRIORITY', 'HIGH_PRIORITY'] %}
+                                {% if dependency_res.status.name == 'OUTDATED_DEPENDENCY' %}
                                     <table class="table">
                                         <thead>
                                             <tr>
@@ -128,7 +128,9 @@
                                                         </div>
                                                     </td>
                                                     <td>
-                                                        <span class="status {{ detail.priority | lower }}">{{ detail.priority }}</span>
+                                                        <div class="table-data__info">
+                                                            <h6>{{ detail.priority }}</h6>
+                                                        </div>
                                                     </td>
                                                     <td>
                                                         <div class="table-data__info">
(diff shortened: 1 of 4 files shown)


badge_server/templates/one-badge.html
```diff
@@ -108,7 +108,7 @@ <h3 class="title-3 m-b-30">
                                 <i class="zmdi zmdi-account-calendar"></i>Dependency Check Result: <span class="status {{ dependency_res.status.name | lower }}">{{ dependency_res.status.value }}</span>
                             </h3>
                             <div class="table-responsive table-data">
-                                {% if dependency_res.status.name in ['LOW_PRIORITY', 'HIGH_PRIORITY'] %}
+                                {% if dependency_res.status.name == 'OUTDATED_DEPENDENCY' %}
```
ylil93:
```suggestion
                                {% if dependency_res.status.name in ['OUTDATED_DEPENDENCY', 'OBSOLETE_DEPENDENCY'] %}
```

badge_server/templates/one-badge.html
```diff
@@ -141,29 +143,21 @@ <h6>{{ detail.latest_version }}</h6>
...
                                     </table>
-                                {% elif dependency_res.status.name == 'UP_TO_DATE' %}
+                                {% elif dependency_res.status.name == 'SUCCESS' %}
                                     <h4 class="title-3 m-b-30">All dependencies are up to date!</h4>
+                                {% elif dependency_res.status.name == 'OBSOLETE_DEPENDENCY' %}
+                                    <h4 class="title-3 m-b-30">{{ dependency_res.deprecated_deps }}</h4>
                                 {% endif %}
                             </div>
                         </div>
-                        {% if dependency_res.deprecated_deps | length > 0 %}
```
ylil93: I think this should not be deleted

badge_server/templates/one-badge.html
liyanhui1228: I thought the new status `OBSOLETE_DEPENDENCY` is the same as deprecated dependencies, is that correct?

badge_server/templates/one-badge.html
liyanhui1228: It used to be separated in two boxes in the UI previously as we didn't have `OBSOLETE_DEPENDENCY` status. And for now it would be strange to place the deprecated dependencies in a separate box from the dependency result box. So that I merged those two.

But if the obsolete dependency status means something else, it makes sense to keep them separated.

badge_server/templates/one-badge.html
liyanhui1228: And the deprecated dependencies is displayed as a list, while the outdated dependency is displayed as a table, so that I let them in two different if-else branches.

ylil93: It's not. Please take a look at this for the old to new mapping:
https://github.com/GoogleCloudPlatform/cloud-opensource-python/blob/70b41cadaf7532cf03cee2e0bcc16f2706c0c5b7/badge_server/main.py#L133

badge_server/templates/one-badge.html
ylil93: `OBSOLETE_DEPEDENCY` replaced `HIGH_PRIORITY` which existed in the previous UI...

badge_server/templates/one-badge.html
liyanhui1228: Acknowledged, thx!

badge_server/templates/one-badge.html
liyanhui1228: Changed.

badge_server/test_badges.py
```diff
@@ -595,3 +666,31 @@ def test_github_py2py3_py3_incompatible_fresh_nodeps(self):
         self.assertEqual(json_response['right_text'], 'self incompatible')
         self.assertEqual(json_response['right_color'], '#E05D44')
         self.assertLinkUrl(package_name, json_response['whole_link'])
+
+
+class TestDependencyOutdatedOrObsolete(BadgeImageDependencyIssueTestCase):
```
ylil93: I actually have an open PR that will add testing on Outdated/Obsolete dependencies, so can you a `# TODO: merge with future dependency test classes once PRs are merged`?

liyanhui1228: Just saw your pending PR, just use yours is fine.

PR #312, opened 2019-05-13
add a framework for badge details page testing

ylil93:
- create a test path for '/one_badge_target' that returns a dict of data that would be applied to html template in production
- add assertBadgeStatusToColor() helper test function
- rename BadgeImageTestCase->BadgeTestCase

### DIFF
--- a/badge_server/main.py
+++ b/badge_server/main.py
@@ -53,7 +53,7 @@ app = flask.Flask(__name__)
 
 
 @enum.unique
-class BadgeStatus(enum.Enum):
+class BadgeStatus(str, enum.Enum):
     """Represents a package's badge status.
 
     The status is based on the results of running 'pip install' and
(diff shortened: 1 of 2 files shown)


badge_server/main.py
```diff
@@ -53,7 +53,7 @@
 
 
 @enum.unique
-class BadgeStatus(enum.Enum):
+class BadgeStatus(str, enum.Enum):
```
liyanhui1228: Why inherit from `str` in here?

ylil93: To make `BadgeStatus` serializable when calling `jsonify()` on a dict containing that object:
https://github.com/GoogleCloudPlatform/cloud-opensource-python/blob/bc21101769dd45744c88c04cb82124d875c036d3/badge_server/main.py#L548

I'm not sure if this is the best way to do this, but it was a popular suggestion when I was looking for ways to do this.

liyanhui1228: This LGTM!

PR #314, opened 2019-05-14
refactor badge image testing

ylil93:
- move all image related assertions to new `assertImageResponse()` method, originating in `BadgeTestCase`

### DIFF
--- a/badge_server/test_badges.py
+++ b/badge_server/test_badges.py
@@ -762,6 +762,12 @@ class TestSelfIncompatible(BadgeTestCase):
 class TestBadgeImageOutdatedDependency(BadgeTestCase):
     """Tests for cases where the badge image displays 'old dependency.'"""
 
+    def assertImageResponse(self, package_name, left_text):
+        """Assert that the badge image response is correct for a package."""
+        status = main.BadgeStatus.OUTDATED_DEPENDENCY
+        BadgeTestCase.assertImageResponse(
+            self, package_name, status, left_text)
+
     def test_pypi_py2py3_off_by_minor(self):
         old_dep_info = dict(UP_TO_DATE_DEPS)
         old_dep_info['google-auth'] = {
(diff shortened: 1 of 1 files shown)


badge_server/test_badges.py
```diff
@@ -548,70 +557,45 @@ def assertTargetResponse(self, package_name, *supported_pyversions):
             json_response['dependency_res'],
             {'deprecated_deps': '', 'details': {}, 'status': expected_status})
 
+    def assertImageResponse(self, package_name, left_text):
```
brianquinlan: How about `assertImageSuccess()` to not override the base class that has a different signature. Likewise for the following classes e.g. `assertImageMissingData()`

I might even have one method for github and one for pypi but that's up to you.

ylil93: I think both suggestions make sense.

badge_server/test_badges.py
```diff
@@ -522,6 +542,16 @@ def assertBadgeStatusToColor(self, badge_status_to_color):
 class TestSuccess(BadgeTestCase):
     """Tests for the cases where the badge image displays 'success.'"""
 
+    def assertImageResponsePyPI(self, package_name):
```
brianquinlan: I'd prefer not to bask a base class method with one with a different signature.

ylil93: I thought that the function names were getting long/convoluted, especially since by looking at the test suite each method is being implemented in, it should be clear what the expected status would be.
(eg `assertImageObsoleteDependencyPyPI()` and `assertImageSelfIncompatibleGithub()`) 

What if I pull `_assertImageResponsePyPI()` / `_assertImageResponseGithub()` from the base class and have other classes implement `assertImageResponsePyPI()` / `assertImageResponseGithub()`?

brianquinlan: OK, let's just leave it the way that it is then.

PR #316, opened 2019-05-21
[review requested] add details page testing

ylil93:
Please review
- covers unknown package, missing data, self incompatible cases
- todo: add details page testing for dependency tests, add test suite for pair incompatibility


ylil93:
I realized the current nox configuration *does* use the local code to install compatiblity_lib for testing and that the failing test case was in `test_badge_server.py` not `test_badges.py`

PR #321, opened 2019-05-24
update self incompatible tests to represent the data model

ylil93:
- add GOOGLE_API_CORE_SELF_INCOMPATIBLE_DATA
- update self incompatible tests to represent the data model

### DIFF
--- a/badge_server/test_badges.py
+++ b/badge_server/test_badges.py
@@ -955,85 +955,102 @@ class TestSelfIncompatible(BadgeTestCase):
         BadgeTestCase._assertImageResponseGithub(
             self, package_name, main.BadgeStatus.SELF_INCOMPATIBLE)
 
-    def assertTargetResponse(self, package_name, *affected_pyversions):
+    def assertTargetResponse(self, package_name, expected_pair_result):
         json_response = self.get_target_json(package_name)
         self.assertEqual(json_response['package_name'], package_name)
         self.assertBadgeStatusToColor(json_response['badge_status_to_color'])
 
         # self compatibility result check
-        for pyversion in ['py2', 'py3']:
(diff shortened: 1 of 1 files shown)


badge_server/test_badges.py
```diff
@@ -250,6 +250,13 @@
     status=compatibility_store.Status.SUCCESS,
     timestamp=datetime.datetime(2019, 5, 7, 0, 0, 0))
 
+APACHE_BEAM_GOOGLE_API_CORE_RECENT_PAIR_INCOMPATIBLE_2 = compatibility_store.CompatibilityResult(
+    [package.Package('apache-beam[gcp]'),
```
brianquinlan: Aren't there messages for these?

ylil93: What do you mean by messages?

ylil93:
> It would be better, I think, if I got to see how these constants are used.

I'm thinking for the self incompatibility tests, instead of whatever we're currently saving to the fake store, we use this list instead, so something like:

```
package_name = 'google-api-core'
self.fake_store.save_compatibility_statuses(GOOGLE_API_CORE_SELF_INCOMPATIBLE_DATA)
self.assertImageResponsePyPI(package_name)
self.assertTargetResponse(...)
```

badge_server/test_badges.py
brianquinlan: Like the message that explains the error? I guess we aren't setting those anywhere else because they are only relevant for the details page.

ylil93:
> I mean, could you include your usage of these constants in this PR? Or would that be problematic?

Yeah I can include those changes too in this PR.

badge_server/test_badges.py
brianquinlan: OK, you can leave it in this PR I guess, but you should add them for the details tests.

PR #324, opened 2019-05-24
Fix a bug where installation failures were considered internal errors

brianquinlan:


badge_server/test_badges.py
```diff
@@ -855,6 +873,15 @@ def test_pypi_py2py3_py2_incompatible_fresh_nodeps(self):
         self.assertImageResponsePyPI(package_name)
         self.assertTargetResponse(package_name, 'py2')
 
+    def test_pypi_py2py3_py2_installation_failure_fresh_nodeps(self):
```
ylil93: Don't you also need to add a test for the github case as well as tests for py3?

brianquinlan: Done. I think that your fixes to TestSelfIncompatible will also require fixes to my tests i.e. because all pairwise install combinations will also fail.

ylil93: Ok maybe you can add the py3 tests after I merge those fixes in.

PR #327, opened 2019-06-05
update package and whitelists with new google python packages

ylil93:


compatibility_lib/compatibility_lib/configs.py
```diff
@@ -47,34 +49,44 @@ def _format_url(repo_name, setuppy_path=''):
     'google-cloud-automl',
     'google-cloud-bigquery',
     'google-cloud-bigquery-datatransfer',
+    'google-cloud-bigquery-storage',
```
liyanhui1228: Are those packages newly released to PyPI?

ylil93: I commented out the ones that haven't been released yet, but seems like update_dashboard tests are failing, so it might be that I missed some? I will check it out.

liyanhui1228: Yeah it happened before when a new package exists in the google-cloud-python github repo but hasn't been released to PyPI yet, which will cause our server to fail.

And per offline discussion, skip the tests until the data for newly added packaged populated in Cloud SQL makes sense.

PR #332, opened 2019-06-10
bump compatibility server version in deployment file

ylil93:


compatibility_server/deployment/deploy.yaml
```diff
@@ -29,7 +29,7 @@ spec:
         run: compatibility-server
     spec:
       containers:
-      - image: gcr.io/python-compatibility-tools/compatibility-server:v24
+      - image: gcr.io/python-compatibility-tools/compatibility-server:v29
```
liyanhui1228: I saw in our cloud project the current running image is `compatibility-image:v1`, are you plan to change back the image name to `compatibility-server`?

ylil93: I redeployed a few hours ago to v29.

liyanhui1228: That will be necessary to keep the name consistent. Like in [other places](https://github.com/GoogleCloudPlatform/cloud-opensource-python/blob/master/compatibility_server/run-in-docker.sh#L21) it's built as `compatibility-image` but not `compatibility-server`.

ylil93: I'm not sure how the fork in naming happened, but I changed it to `compatibility-image` in my last commit.

liyanhui1228: And I looked into the `compatibility-image` [versions](https://pantheon.corp.google.com/gcr/images/python-compatibility-tools/GLOBAL/compatibility-image?project=python-compatibility-tools&folder&organizationId=433637338589&gcrImageListsize=30). Currently there is a v1 and a v29, would look better if the version numbers are consecutive.

ylil93: Last week I mistakenly set the version to v1 when I redeployed the compatibility server then. When I checked the history, the last version before that was v28, so I redeployed a few hours ago to correct the error.

liyanhui1228: The image name after `docker build` command will be the image name you deployed. Though that run-in-docker.sh script looks like is for local testing. And I'm fine with using either of the image names, we just need to be consistent across the code and docs. Like if we switch to use `compatibility-image` the README file for deployment guide commands will need to be updated.

PR #342, opened 2019-07-01
create a manifest file for compatibility service

ylil93:
for establishing a static IP address according to:
https://cloud.google.com/kubernetes-engine/docs/tutorials/configuring-domain-name-static-ip#step_3_visit_your_reserved_static_ip_addres


compatibility_server/service/compatibility-service.yaml
```diff
@@ -0,0 +1,15 @@
...
+    app: compatibility-server
+spec:
+  selector:
+    app: compatibility-server
+    tier: web
+  ports:
+  - port: 80
+    targetPort: 8080
+  type: LoadBalancer
+  loadBalancerIP: 35.184.23.203
```
brianquinlan: Should this be parameterized some how? Because presumably it only works for one cluster.

ylil93: I think we only have a single service though? And it would be something externally facing:
https://cloud.google.com/kubernetes-engine/docs/tutorials/configuring-domain-name-static-ip#step_3_visit_your_reserved_static_ip_address

compatibility_server/deployment/service.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: compatibility-server
```
brianquinlan: Can you change the name? Like `compatibility-server-service`

ylil93: done

compatibility_server/deployment/service.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: compatibility-server-service
```
brianquinlan: Did you tests that this works?

ylil93: yep

PR #355, opened 2020-01-24
chore(compatibility_server): extend timeout on requests

orthros:
This PR extends the timeout of the `compatibility_server` to 400 seconds (from 300).

This fixes recent timeout errors.

Signed-off-by: Colin Nelson <colnnelson@google.com>

### DIFF
--- a/compatibility_server/Dockerfile
+++ b/compatibility_server/Dockerfile
@@ -34,5 +34,5 @@ EXPOSE 8888
 # 5 replicas per vm, each replica has 10 workers which allows 50 docker images
 # created at most in the vm.
 CMD cd compatibility_checker && \
-  gunicorn -b 0.0.0.0:8888 -w 2 --worker-class sync --max-requests 20 --max-requests-jitter 10 --timeout 300 \
+  gunicorn -b 0.0.0.0:8888 -w 2 --worker-class sync --max-requests 20 --max-requests-jitter 10 --timeout 400 \
            -e EXPORT_METRICS=1 compatibility_checker_server:app


compatibility_server/Dockerfile
```diff
@@ -34,5 +34,5 @@ EXPOSE 8888
 # 5 replicas per vm, each replica has 10 workers which allows 50 docker images
 # created at most in the vm.
 CMD cd compatibility_checker && \
-  gunicorn -b 0.0.0.0:8888 -w 2 --worker-class sync --max-requests 20 --max-requests-jitter 10 --timeout 300 \
+  gunicorn -b 0.0.0.0:8888 -w 2 --worker-class sync --max-requests 20 --max-requests-jitter 10 --timeout 400 \
```
brianquinlan: You could make this change and redeploy to see if it changes anything but it looks like there were ~300 timeouts for >3200 successful checks. It could be that you need more machine resources or something. Did the whitelist change recently?

orthros: As far as I can tell, the whitelist hasn't been changed. The cluster looks healthy, but that's a bit difficult to determine because of  the DIND setup.

I can scale the cluster a bit higher on machine resources to see, but that'll require updating the deployment specification as well.

brianquinlan: You can try increasing the timeout and seeing what happens...

orthros: Follow up: I don't have permissions to update the cluster. Its currently sitting at 200 nodes (matching the deployment spec & ant-affinity policy)

brianquinlan: I made you an admin of the project. I think that you can make your permissions whatever you want now.

orthros:
Closed as resolved with Node Upgrades in the Cluster

